PR #198, opened 2026-06-17
fix: pass `GITHUB_TOKEN` to `Repair-WinGetPackageManager`

UnownPlain:
`Repair-WinGetPackageManager` automatically authenticates its requests to GitHub's API if a `GITHUB_TOKEN` environment variable is present (microsoft/winget-cli#6071), which should fix some transient errors in CI. This also fixes some findings from https://zizmor.sh/ and changes the `detect` job's runner to `ubuntu-slim` since a full runner isn't necessary for that job.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -4,25 +4,29 @@ on:
   workflow_dispatch:
     inputs:
       package_id:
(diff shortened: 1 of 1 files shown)


pl4nty:
Thanks, I was wondering why `Repair-WinGetPackageManager` was failing intermittently

PR #199, opened 2026-06-18
fix: install latest WinGet pre-release from GitHub

UnownPlain:
microsoft/winget-cli#6071 hasn't actually landed in a release yet so `Repair-WinGetPackageManager` still sends unauthenticated requests. I updated the script to fetch via GitHub releases as a workaround. Also noticed that arm64 test runs were hitting the issue fixed in microsoft/winget-cli#6127, so this PR also switches to using the latest pre-release which fixes it.


pl4nty:
Thanks! tbh I wasn't sure about using `Repair-WinGetPackageManager` in the first place, downloading from GitHub releases seems simpler

PR #208, opened 2026-06-20
fix: use env vars to prevent template injection

UnownPlain:
See https://docs.zizmor.sh/audits/#template-injection.


pl4nty:
thanks!

PR #233, opened 2026-06-24
New package: Microsoft.DotNet.Native.Framework version 2.2.29512

DandelionSprout:
Checklist for Pull Requests

  - Relates to https://github.com/microsoft/winget-pkgs/pull/365045.

Manifests

- [x] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [x] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.12 schema](https://github.com/microsoft/winget-pkgs/tree/master/doc/manifest/schema/1.12.0)?

---

### DIFF
--- a/manifests/m/Microsoft/DotNet/Native/Framework/2.2.29512/Microsoft.DotNet.Native.Framework.installer.yaml
+++ b/manifests/m/Microsoft/DotNet/Native/Framework/2.2.29512/Microsoft.DotNet.Native.Framework.installer.yaml
@@ -1,0 +1,31 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: Microsoft.DotNet.Native.Framework
+PackageVersion: 2.2.29512
+Platform:
+- Windows.Universal
+MinimumOSVersion: 10.0.10049.0
+InstallerType: zip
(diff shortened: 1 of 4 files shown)


pl4nty:
Thanks, this is a good one to have. I reckon rename it to Microsoft.DotNet.Native.Framework.2.2 for consistency, even though .NET Native is really deprecated and shouldn't get new versions

DandelionSprout:
The .appx files' internal file versions are 2.2.29512.0. At least over at pkgs, that would have left me with no other options than to use those internal file versions, or else their validations would break.

pl4nty:
Yeah, I'm thinking `Microsoft.DotNet.Native.Framework.2.2` version 2.2.29512. Because technically it can be installed side-by-side with minor versions like `Microsoft.NET.Native.Framework.2.1` 2.1.27427.0

DandelionSprout:
Thinking further about it, since this is after all the newest version of Native Framework altogether, surely the 2.2.x version doesn't need further sub-folders, whereas any earlier versions like Native Framework 1.7 would?

pl4nty:
Hmm that should be ok, as long as they don't release new major or minor versions

PR #323, opened 2026-07-09
fix: use JSON mode for tj-actions/changed-files

### LINKED ISSUE #310: The build matrix fails to handle PackageVersion-s with spaces in them.
Based on https://github.com/pl4nty/winget-extras/actions/runs/28777887865/job/85326096245.

The package in question has the PackageVersion `R610 U1`. The build matrix seemingly fails to wrap a filepath correctly, leading to it cutting off the path after simply `R610`, and then fails due to not being able to find the non-existent `R610` filepath.

UnownPlain:
Resolves #310

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -23,6 +23,8 @@ jobs:
           files: |
             manifests/**/*.installer.yaml
             fonts/**/*.installer.yaml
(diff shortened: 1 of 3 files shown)


pl4nty:
Thanks!

PR #324, opened 2026-07-09
Use GitHub API for authenticated dependency downloads

pl4nty:
## Description

This change replaces the unauthenticated `raw.githubusercontent.com` download method with the authenticated GitHub API (`gh`) to avoid rate limiting issues (HTTP 429 errors).

### Changes

- Replaced `curl` with `gh api` for downloading dependency files
- Uses the GitHub API's raw content endpoint with proper authentication
- Removes the parallel curl approach in favor of sequential API calls
- Eliminates rate limiting issues by using authenticated requests

### Rationale

The previous implementation downloaded files from `raw.githubusercontent.com`, which is unauthenticated and subject to strict rate limits. This caused failures when downloading multiple dependencies. The GitHub API provides authenticated access with higher rate limits, making the workflow more reliable.

### Testing

The merge workflow will be tested during the next scheduled run or manual trigger. The change maintains the same functionality while improving reliability through authenticated API access.

https://claude.ai/code/session_01FuhaPaPHXpV2SCzKAg1A5h

### DIFF
--- a/.github/workflows/merge.ps1
+++ b/.github/workflows/merge.ps1
@@ -86,9 +86,12 @@ if ($missingDependencies) {
     $destination = Join-Path $using:tempDependencies $path
     New-Item $destination -ItemType Directory -Force | Out-Null
 
+    $curlArguments = @('--fail', '--silent', '--show-error', '--location', '--parallel')
     foreach ($file in $files) {
-      & gh api $file.url --header 'Accept: application/vnd.github.raw' > (Join-Path $destination $file.name)
+      $uri = "https://cdn.jsdelivr.net/gh/microsoft/winget-pkgs@master/$(& $encodePath "$path/$($file.name)")"
+      $curlArguments += @('--output', (Join-Path $destination $file.name), $uri)
     }
+    & curl @curlArguments
   }
 }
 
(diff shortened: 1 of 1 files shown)


pl4nty:
This could be much slower

UnownPlain:
Fetching from `cdn.jsdelivr.net` might be a better option as they appear to not have rate limits. Anthelion uses it to quickly check if a version already exists in the repo.

pl4nty:
`cdn.jsdelivr.net` [403'd somehow](https://github.com/pl4nty/winget-extras/actions/runs/28986081402/job/86015396711), then succeeded on a rerun that just added more logging. Going to merge this for now and keep an eye on it

UnownPlain:
My best guess is that it tried to fetch a ton files that weren't cached on their CDN and had it had to fetch them from GitHub. I ran it 100 times with `hyperfine` on my fork and it appears to work fine.

```
Benchmark 1: pwsh -NoLogo -NoProfile -File ./.github/workflows/merge.ps1
 
  Warning: The first benchmarking run for this command was significantly slower than the rest (8.682 s). This could be caused by (filesystem) caches that were not filled until after the first run. You should consider using the '--warmup' option to fill those caches before the actual benchmark. Alternatively, use the '--prepare' option to clear the caches before each timing run.
  Time (mean ± σ):      2.085 s ±  0.667 s    [User: 2.783 s, System: 0.960 s]
  Range (min … max):    1.935 s …  8.682 s    100 runs
```

Anthelion also sends hundreds of requests (1 per shard if the version fetch was successful) to jsdelivr over a few seconds every 7 minutes and isn't experiencing issues.

PR #342, opened 2026-07-13
chore: remove UTF-8 BOM from manifests

UnownPlain:
Don't think we should have BOM in this repo as a lot of tools expect UTF-8 with no BOM. CI should probably also check for this.


pl4nty:
thanks! agreed, CI should lint this and the LSP header

DandelionSprout:
I started using BOM in pretty much all of my projects ever since https://github.com/AdguardTeam/FiltersRegistry/issues/1142, so I'm still in the process of adapting to that a whole lot of others' projects specifically don't want BOM.

PR #376, opened 2026-07-17
fix(validation): install WinGet dependencies

UnownPlain:
This should fix the dependency errors seen recently.


pl4nty:
Thanks! Good to be back on a stable release too

PR #387, opened 2026-07-19
chore: remove architecture suffixes from commands

UnownPlain:
I think most people would prefer to not have the architecture at the end of commands.


pl4nty:
Sounds good, I think architecture would only make sense for side-by-side installs

PR #431, opened 2026-07-20
be5invis.Iosevka: correct InstallerUrl

UnownPlain:


pl4nty:
oh good catch, thanks!

PR #432, opened 2026-07-20
subframe7536.MapleMono.*: add other variants

UnownPlain:


UnownPlain:
Looks like CI failures require changes in komac so non-default branches work. Give me some time to fix this

PR #449, opened 2026-07-21
feat(shard/test): support PRs from forks

UnownPlain:
Should fix these failures: https://github.com/pl4nty/winget-extras/actions/runs/29787727191?pr=432


pl4nty:
Thanks for this! Can't find a way to dispatch the test workflow on your branch, so we'll just have to test later

PR #414, opened 2026-07-20
chore: correct some folder paths

UnownPlain:
I had some extra tokens to burn since Codex rate-limits keep getting reset, so I extended the lint-manifests script to catch manifest and path errors. Curious whether you think it would be a good idea to add these checks to CI.

https://github.com/pl4nty/winget-extras/compare/main...UnownPlain:winget-extras:manifest-validation

```
$ bun scripts/lint-manifests.ts
error[lint-manifests] --> fonts/m/Microsoft/FluentUISystemIcons/Light/1.1.331/Microsoft.FluentUISystemIcons.Filled.installer.yaml:3:1
  2 | 
> 3 | PackageIdentifier: Microsoft.FluentUISystemIcons.Light
>   | ^^^^^^^^^^^^^^^^^ manifest path does not match PackageIdentifier
  4 | PackageVersion: 1.1.331
  5 | ReleaseDate: 2026-06-26
  help: rename the file to Microsoft.FluentUISystemIcons.Light.installer.yaml

error[lint-manifests] --> fonts/m/Microsoft/FluentUISystemIcons/Light/1.1.331/Microsoft.FluentUISystemIcons.Filled.locale.en.yaml:3:1
  2 | 
> 3 | PackageIdentifier: Microsoft.FluentUISystemIcons.Light
>   | ^^^^^^^^^^^^^^^^^ manifest path does not match PackageIdentifier
  4 | PackageVersion: 1.1.331
  5 | PackageLocale: en
  help: rename the file to Microsoft.FluentUISystemIcons.Light.locale.en.yaml

error[lint-manifests] --> fonts/m/Microsoft/FluentUISystemIcons/Light/1.1.331/Microsoft.FluentUISystemIcons.Filled.yaml:3:1
  2 | 
> 3 | PackageIdentifier: Microsoft.FluentUISystemIcons.Light
>   | ^^^^^^^^^^^^^^^^^ manifest path does not match PackageIdentifier
  4 | PackageVersion: 1.1.331
  5 | DefaultLocale: en
  help: rename the file to Microsoft.FluentUISystemIcons.Light.yaml

error[lint-manifests] --> manifests/b/bol-van/zapret2/1.0.2/bol-van.Zapret2.installer.yaml:3:1
  2 | 
> 3 | PackageIdentifier: bol-van.zapret2
>   | ^^^^^^^^^^^^^^^^^ manifest path does not match PackageIdentifier
  4 | PackageVersion: 1.0.2
[...]

PR #459, opened 2026-07-22
feat: manifest validation

UnownPlain:
This PR implements the checks mentioned in #414. I know this adds a lot of code but I think it's worth it to ensure the consistency and correctness of manifests in this repo.

Here is a list of what this enforces (mostly written by Codex):

### YAML and schema validity

- Manifests must contain valid YAML.
- Duplicate mapping keys are rejected.
- Complex YAML constructs fall back to the stricter parser so duplicate keys and syntax errors are not missed.
  - [Bun.YAML](https://bun.com/docs/runtime/yaml) is preferred over [yaml](https://npmx.dev/package/yaml) for performance reasons.
- Only manifest schema versions `1.9.0`, `1.10.0`, `1.12.0`, and `1.28.0` are supported.
- Supported manifest types are:
  - `version`
  - `installer`
  - `defaultLocale`
  - `locale`
- [Manifests are validated against](https://redirect.github.com/repology/repology-updater/issues/1392#issuecomment-2144956402) the [corresponding official WinGet JSON schema](https://redirect.github.com/microsoft/winget-pkgs/issues/161793).
- This includes:
  - Required properties.
  - Property types.
  - Allowed enum values.
  - String patterns and length limits.
  - Array size and uniqueness constraints.
  - Object structure.
  - [Rejection](https://github.com/microsoft/winget-pkgs/pulls?q=in%3Atitle%20Fix%20typo%20in%20manifest%20key) of [unknown properties.](https://github.com/microsoft/winget-pkgs/blob/809fef993d44b61713197f81a4bb29b94fb8ed09/manifests/m/Minitab/MinitabStatisticalSoftware/22/22.4.0.0/Minitab.MinitabStatisticalSoftware.22.locale.en-US.yaml#L10-L18)
  - Installer URL and SHA-256 formatting.
  - Package identifier, version, locale, and architecture formatting.

The schemas are tightened and additional restrictions are applied:

- `ManifestVersion` must exactly match the schema version being used.
- `Channel` is not supported and must be absent or null.
- Approximate `PackageVersion` and `DisplayVersion` values beginning with `< ` or `> ` are rejected.
- Windows feature dependency names may contain only letters, digits, underscores, and hyphens.
- Installer switches cannot contain UNC/network-share paths using double backslashes.
- Authentication modes other than `none` are rejected.
- PowerShell DSC resources are rejected.
- If no root `InstallerType` exists, every installer entry must define one.
- Every agreement entry must contain at least one nonblank `AgreementLabel`, `Agreement`, or `AgreementUrl`.
- A default-locale manifest must contain `Moniker`.

### Filesystem requirements

- [Manifest YAML files must use UTF-8 without a BOM.](https://redirect.github.com/microsoft/winget-pkgs/issues/1863)
- YAML paths must be regular files; symlinks and special files are rejected.
- [Every regular file under the manifest and font roots must have mode `0644`.](https://redirect.github.com/microsoft/winget-pkgs/issues/90127)
- Non-UTF-8 files can be converted automatically from:
  - UTF-8 with BOM.
  - UTF-16 little-endian.
  - UTF-16 big-endian.
  - Windows-1252.

### Repository layout

- Package directories must follow:

  ```text
  <root>/<lowercase first character>/<PackageIdentifier segments>/<PackageVersion>
  ```

- The first-character directory must be lowercase.
- Package identifier segments must preserve the exact identifier casing.
- Sibling directories cannot differ only by case.
- Identifier casing must remain consistent across every version of a package.
- Installer manifests must be named:

  ```text
  <PackageIdentifier>.installer.yaml
  ```

- Version manifests must be named:

  ```text
  <PackageIdentifier>.yaml
  ```

- Locale and default-locale manifests must be named:

  ```text
  <PackageIdentifier>.locale.<PackageLocale>.yaml
  ```

- Manifest directories may contain only regular `.yaml` files.
- Subdirectories are not allowed inside version directories.
- Non-manifest files are allowed only at the package level, immediately above version directories.
- Nonempty directories that do not belong to a valid package path are rejected.
- Empty unrelated directories are ignored.

### Font and application separation

- Font installers must be under the font root.
- Application installers must be under the application manifest root.
- A package identifier cannot exist under both roots.
- A package cannot mix font and application installer types across its versions.

### Manifest-set consistency

Every package-version directory must contain exactly:

- One version manifest.
- One installer manifest.
- One default-locale manifest.
- Zero or more additional locale manifests.

The linter also requires:

- Every manifest to have the same `PackageIdentifier`.
- Every manifest to have the same `PackageVersion`.
- Every manifest to have the same `ManifestVersion`.
- Every localization to have a unique `PackageLocale`.
- The version manifest’s `DefaultLocale` to match the `PackageLocale` of the default-locale manifest.

### Schema headers

- Every manifest must contain exactly one YAML language-server schema header.
- The header must exactly match the manifest type and version.
- A single incorrect header can be fixed automatically.
- Missing or duplicate headers require manual correction.

### Installer inheritance

Installer-level validation runs against resolved values:

- Root installer properties are inherited where WinGet would inherit them.
- Installer-level non-null values override root values.
- Installer switches are merged individually.
- ZIP nested-installer metadata is resolved separately.
- Root `AppsAndFeaturesEntries`, `PackageFamilyName`, `ProductCode`, and dependencies are inherited only for compatible installer types.

### Duplicate installers

Installer entries are considered duplicates when they have the same:

- Normalized installer type, treating `appx` as `msix`.
- Nested installer type for ZIP installers.
- Architecture.
- Installer locale.
- Overlapping scope.

An unspecified scope overlaps both user and machine scope.

Only one duplicate-installer diagnostic is emitted per installer manifest.

### Installer URL and hash consistency

- One `InstallerUrl` cannot be associated with multiple SHA-256 hashes.
- Reusing one SHA-256 hash for multiple URLs produces a warning.
- Hash comparisons are case-insensitive.

### Installer identity metadata

- `PackageFamilyName` on a non-MSIX installer produces a warning unless an Apps and Features entry identifies the installer as MSIX.
- `ProductCode` is accepted only for:
  - `exe`
  - `inno`
  - `msi`
  - `nullsoft`
  - `wix`
  - `burn`
  - `portable`
- `AppsAndFeaturesEntries` is accepted only for those same effective installer types.

### Portable installers

- A directly declared portable installer may expose at most one command.
- An effective portable installer may contain at most one Apps and Features entry.
- Declaring `Scope` for a portable installer produces a warning.

### ZIP installers

- ZIP installers must define `NestedInstallerType`.
- ZIP installers must define at least one `NestedInstallerFiles` entry.
- Non-portable and non-font ZIP installers may contain only one nested installer.
- `RelativeFilePath` cannot escape the archive through `..`.
- `PortableCommandAlias` cannot escape its base directory.
- Portable command aliases must be unique case-insensitively.
- Duplicate path-and-alias pairs are rejected case-insensitively.
- The same nested path is currently allowed when it exposes different aliases.

Font archives accept these extensions:

- `.otf`
- `.ttf`
- `.fnt`
- `.ttc`
- `.otc`

### Installer return codes

- Signed return codes are normalized to their equivalent unsigned 32-bit representation.
- A return code cannot appear both in `InstallerSuccessCodes` and `ExpectedReturnCodes`.
- Duplicate expected return codes are rejected.
- Schema-level array uniqueness also rejects duplicate success-code entries.

### MSI and WiX switches

For effective MSI and WiX installers:

- Only supported MSI quiet, logging, and property syntax is accepted.
- Long options such as `quiet`, `passive`, `norestart`, `forcerestart`, `promptrestart`, and `log` are normalized before validation.
- Quiet-mode modifiers are validated.
- Log modifiers must be valid and include a log filename.
- Quoting and MSI property syntax are validated.
- These MSI properties are blocked:
  - `TRANSFORMS`
  - `PATCH`
  - `MSINEWINSTANCE`
  - `ADMINPROPERTIES`

### Warning behavior

The linter currently produces warnings for:

- `PackageFamilyName` on an unusual installer type.
- `Scope` on a portable installer.
- One SHA-256 hash being reused by multiple URLs.

Warnings do not fail the command unless `--deny-warnings` is supplied.

### Automatic fixes

Automatic fixes are available for:

- UTF-8 BOM removal.
- Supported legacy encoding conversion to UTF-8.
- File mode correction to `0644`.
- Replacement of one incorrect schema header.

## [`winget validate`](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCLICore/Commands/ValidateCommand.cpp) rules not implemented

Compared with the current upstream full-validation path in [`winget validate`](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCLICore/Commands/ValidateCommand.cpp) and its [semantic validator](https://github.com/microsoft/winget-cli/blob/master/src/AppInstallerCommonCore/Manifest/ManifestValidation.cpp), these checks are not implemented or do not have full parity:

- **Complete WinGet version parsing.**  
  WinGet parses `PackageVersion` and Apps and Features `DisplayVersion` using its version implementation and rejects values it cannot parse. The linter currently applies schema patterns and the explicit approximate-version restriction, but does not reproduce the complete WinGet version parser.

- **Native Windows URL validation.**  
  WinGet validates installer URLs using the Windows URL parser. The linter enforces the schema’s HTTP/HTTPS pattern but may accept malformed URLs that Windows rejects.

- **Complete BCP 47 validation.**  
  WinGet validates `PackageLocale` and `InstallerLocale` with its locale implementation. The schema regex used by the linter is less precise and may accept syntactically shaped but invalid BCP 47 tags.

- **Missing silent switches for EXE installers.**  
  WinGet warns when an effective `exe` installer lacks either `Silent` or `SilentWithProgress`. The linter does not currently emit this warning.
  - Some manifests do not require switches install silently, so this linter doesn't check this.

- **Portable ZIP file type.**  
  WinGet requires nested portable installers to reference `.exe` files. The linter validates font extensions but does not restrict nested portable files to `.exe`.
  - We have packages with scripts and allows scripts unlike the community repo.

- **Network URLs inside installer switches.**  
  WinGet warns when switches contain `http://`, `https://`, or `ftp://`. The linter rejects UNC paths but does not detect these URL forms.
  - We allow network addresses in switches and have packages with network addresses in switches. [The community repo's pipelines have recently started throwing errors for manifests with this](https://dev.azure.com/shine-oss/winget-pkgs/_build/results?buildId=365551&view=logs&j=dc5484e7-078c-5e69-1cbb-0f132d85a80e&t=0ff99015-932f-506f-8afc-34c2cd60b3e2&l=83).

- **Verified-publisher agreement warning.**  
  WinGet warns that use of the `Agreement` text field requires a verified publisher. The linter validates that an agreement has content but does not perform publisher-verification policy checks.
  - We don't require this and the community repo doesn't enforce this either.

- **Duplicate nested paths with different aliases.**  
  WinGet rejects duplicate `RelativeFilePath` values regardless of their `PortableCommandAlias`. The linter currently considers the path-and-alias pair, so the same file can be exposed through multiple distinct aliases.
  - This rule is stupid. This isn't implemented so stuff like the below example isn't required to appease the linter:

```yaml
# https://github.com/microsoft/winget-pkgs/blob/cc5721ad92e992de14c597c4f4e3e038b1d8dd46/manifests/l/Lean/Elan/4.2.3/Lean.Elan.installer.yaml#L8-L22
NestedInstallerFiles:
- RelativeFilePath: elan-init.exe
  PortableCommandAlias: elan
- RelativeFilePath: .\elan-init.exe
  PortableCommandAlias: lake
- RelativeFilePath: .\.\elan-init.exe
  PortableCommandAlias: lean
- RelativeFilePath: .\.\.\elan-init.exe
  PortableCommandAlias: leanc
- RelativeFilePath: .\.\.\.\elan-init.exe
  PortableCommandAlias: leanchecker
- RelativeFilePath: .\.\.\.\.\elan-init.exe
  PortableCommandAlias: leanmake
- RelativeFilePath: .\.\.\.\.\.\elan-init.exe
  PortableCommandAlias: leanpkg
```

- **All supported schema versions and manifest formats.**  
  WinGet supports more schema versions, including newer versions such as `1.30.0`, and can validate singleton or merged manifests. The linter supports only the four selected schema versions and repository-style multi-file manifests.

There are also two behavior differences:

- WinGet treats schema-header problems as warnings during `winget validate`; this linter treats them as errors.
- `winget validate` fails on warnings by default unless `--ignore-warnings` is used. This linter allows warnings by default and fails them only with `--deny-warnings`.


pl4nty:
Great work, thanks! I particularly like the rules-based architecture, I'll probably add more rules in the future.

More broadly, I hope your linter and other ideas in this repo might be seen by winget-pkgs too - and show how they could improve submission UX, updates, etc. I'm very happy to accept PRs even if they're adding a lot of code.

> Network URLs inside installer switches

Looks like this was an undocumented addition in https://github.com/microsoft/winget-cli/pull/6169. I guess it'll break other popular packages like `Microsoft.Office`.

PR #543, opened 2026-07-26
feat(linter): warn on packages without a shard

pl4nty:
Adds a `repository/shard-coverage` rule for packages with no `shards/json/<id>.json` or `shards/script/<id>.ts`, using the `.Font` suffix for fonts and counting `.disabled` shards as covered. Warning-level so `manifests:check` still passes — currently 109 of 289 packages.

It also flags `shards/json/ryanoasis.IosevkaTermSlab.Font.json`, which doesn't match its package `ryanoasis.Iosevka.Term.Slab`.

https://claude.ai/code/session_019n7eTSi21sG4uCvmYahAjD


UnownPlain:
There are some packages that haven't updated in the past decade or are deprecated/abandoned. Probably not worth adding shards for these packages.

pl4nty:
Agreed, I found many when generating shards for winget-pkgs. Not sure how to track them though. Might use `.disabled` for now

PR #544, opened 2026-07-26
feat(linter): require pinned github.com installer URLs

pl4nty:
Adds two rules enforcing one style for GitHub downloads, and fixes every manifest they flag — the linter now reports zero warnings.

- **`installer/github-host`** — bans `raw.githubusercontent.com` and `codeload.github.com`, where github.com's `raw` and `archive` URLs land, so the repository stops mixing both styles. Release assets redirect to a signed, expiring `release-assets.githubusercontent.com` URL that can't appear in a manifest, so most GitHub URLs here must redirect regardless — canonical github.com is the only style every download can share.
- **`installer/github-pinning`** — github.com `raw`/`blob`/`archive` downloads must use `refs/tags/<tag>` or a commit SHA, since a URL can't say whether a bare ref like `1.1.333` is a tag or a branch. Anything else can move out from under `InstallerSha256`.

```
warning[installer/github-pinning] --> fonts/g/Google/Tinos/1.231/Google.Tinos.installer.yaml:15:17
> 15 | InstallerUrl: https://github.com/googlefonts/tinos/archive/refs/heads/main.zip
>    |               ^^^^^^^^ InstallerUrl must use a pinned tag or commit
```

## Manifest fixes

Every rewritten URL's bytes were verified against the recorded `InstallerSha256`; tags are preferred wherever a tag serves those exact bytes (per review):

- `raw.githubusercontent.com` → `github.com/<owner>/<repo>/raw/…` across the flagged manifests
- Version tags: Meslo LG → `v1.2.1`, FluentUI System Icons → `1.1.331`/`1.1.333`/`1.1.334`, Fluent Emoji Webfont → `v0.8.5`, .NET Native Runtime → `app-metadata@v0.1.16`, Scribble → `msix-packaging@MSIX-Core-1.2-release` (Git LFS; the oid is the recorded hash)
- Commit pins where no tag serves the bytes: Monoid and Victor Mono (built zips committed outside/after tags), sunuwar, Shafarik, OpenSans, EgyptianOpenType, Quicksand (no usable tags)
- `shaunsingh.SFMonoLigaturized 3.4.0`: branch archive → `archive/dc5a3e6….zip`. The repo is unchanged since 2023 so the tree is identical; the zip's folder prefix changes, so `InstallerSha256` (computed from the real download) and `NestedInstallerFiles` paths are updated together.
[...]

PR #548, opened 2026-07-26
chore: add more shards

pl4nty:
Adds shards for packages with a stable, declarative update source, brings existing shards in line with the GitHub URL rules from #544, and starts covering packages whose installer URL is pinned but never moves.

Checklist for Pull Requests

  - No related winget-pkgs issue. Builds on #543, and picks up #549 and #544 from main.

Manifests

- The only manifest change is removing `Microsoft.TeamsMeetingAddin` (below). `bun manifests:check` (1253 files), `bun test:manifests` (66 tests) and `bun fmt` pass, and all 233 JSON shards validate against `https://anthelion.unownplain.dev/schema.json`.

## 1. Automated update shards (57 packages)

| Group | Count | Strategy |
| --- | --- | --- |
| IBM Plex families | 16 | `github-release` with `tagFilter` + `{github.rawTag}` |
| Iosevka SuperTTC variants (Aile, Curly, Curly Slab, Etoile, Slab, SS03–SS16) | 15 | `github-release` |
| Nerd Fonts (Caskaydia Cove/Mono, Inconsolata, Iosevka Term Slab) | 4 | `github-release` |
| PlemolJP, PlemolJP HS, PlemolJP NF | 3 | `github-release` |
| Monaspace NF, Monocraft NF, Sudo, JuliaMono, Awesome Terminal Fonts, SketchyBar App Font, Pretendard, CraftOS-PC, Komac, wxWidgets, Plex Media Server Service Wrapper | 11 | `github-release` |
| SIL Kanchenjunga, SIL Kedebideri | 2 | `github-release` against `silnrsi` |
| .NET Desktop Runtime 5 x86 / 8 x64, .NET Runtime 8 x86 | 3 | `json` against the release-metadata channel feeds |
| Fluent UI System Icons Filled / Light | 2 | `sort-versions` over `tags.atom` |
| Visual Studio Code | 1 | `page-match` against the stable update API |

Two existing shards are renamed because their filenames never matched a package identifier, so they were dead files while the packages they covered were reported as uncovered:

- `IBM.PlexSans.Font.json` → `IBM.Plex.Sans.Font.json`
- `ryanoasis.IosevkaTermSlab.Font.json` → `ryanoasis.Iosevka.Term.Slab.Font.json`

### Notes on the non-obvious ones

**Visual Studio Code** needs the build commit for the download URLs, so one regex over the update API response captures both it and the version. `productVersion` is used verbatim, matching the `1.130.0` directory after #549. `version` is a named group rather than positional, because `commit` comes first and a named group still occupies capture slot 1.

The `json` strategy does not work here: `update.code.visualstudio.com` serves every artefact from a version-addressed redirect, and komac collapses those six redirects into two installers, both arm64, losing all x64 and user-scope entries.

**Fluent UI System Icons** stays on `sort-versions` over `tags.atom`. `github-release` fails outright, since the repo publishes tags but no Release objects. `json` against the tags API picks the wrong tag, because it is a monorepo whose newest tag is currently `react-icons-svg-sprite-subsetting-webpack-plugin@0.0.5`, and `json` can neither filter nor sort.

**Plex Media Server Service Wrapper** tags releases as `v1.2.3` while the installer carries a fourth field, so the shard sets `"version": "{version}.0"` and uses `{packageVersion}` for the asset name.

**.NET Desktop Runtime 8 x64** keeps the `|arm64` override used by the existing `Microsoft.DotNet.DesktopRuntime.9.x64` shard, matching what is already in the manifest. Changing it is a manifest fix rather than a shard fix.

**.NET Desktop Runtime 5 x86** targets an out-of-support channel pinned at 5.0.17, so it resolves correctly and never proposes an update.

## 2. Following the GitHub URL rules from #544

Seven shards emitted `raw.githubusercontent.com`, which `installer/github-host` rejects. They now use the `github.com/{owner}/{repo}/raw/...` form the rule requires and the manifests in #544 already use.

Three are **script** shards — `Google.OpenSans`, `Microsoft.EgyptianOpenType` and `rubjo.VictorMono` — so the URL they generate would have been rejected on their next run, not just in review. The other four are the Fluent icon shards, which now emit `raw/refs/tags/{version}/`.

The hardcoded `1.1.` series filter is also dropped from all four Fluent icon shards, so a future 1.2 or 2.0 tag is picked up. The `releases/tag/` anchor still excludes the monorepo's package tags. This is the change offered in review; happy to revert it if you would rather keep the series pinned.

**`Google.OpenSans` is the exception** and stays on `raw.githubusercontent.com`. Its only variable font has square brackets in the filename, and installer URLs are percent-decoded before download: the run logs `OpenSans%5Bwdth%2Cwght%5D.ttf` and then fails on `OpenSans[wdth,wght].ttf`, which `github.com` 404s while `raw.githubusercontent.com` serves both forms. `installer/github-host` is a warning and `manifests:check` does not pass `--deny-warnings`, so this stays green. Worth noting the same URL is in the manifest as of #544, so that package currently records an installer that cannot be downloaded.

## 3. Updating pins

Three font packages were pinned to a bare commit with no upstream tags, so the pin could never move. They get branch-tracking script shards mirroring the existing Tinos and SpaceMono ones: the head commit pins the URL, and the version comes from the font's `name` table.

| Package | Upstream | Manifest | Dry run |
| --- | --- | --- | --- |
| `Google.AtkinsonHyperlegible.Next` | `googlefonts/atkinson-hyperlegible-next` | 2.001 | 2.001 |
| `Google.AtkinsonHyperlegible.Mono` | `googlefonts/atkinson-hyperlegible-next-mono` | 2.001 | 2.001 |
| `Google.Fira.SansExtraCondensed` | `googlefonts/FiraGFVersion` | 4.203 | 4.203 |

`Google.Fira.SansExtraCondensed` is the one whose pin was actually stale — upstream head is `8d9d02bf`, the manifest pinned `7b7927fa`. The font version is unchanged there, so the shard refreshes the pin without inventing a version bump. Per CONTRIBUTING it gets no seeded `version-state` file, because upstream is ahead of WinGet; the other two are seeded with their current head commit.

`Mozilla.Fira.Sans` and `Mozilla.Fira.Sans.Condensed` get `sort-versions` shards over the `mozilla/Fira` tags. I had held these back because all three Fira packages share one source archive, and it was not clear whether an update would pull every family into each package. The dry run settled it: komac keeps the existing nested-file selection rather than re-globbing the archive, so `Mozilla.Fira.Sans` came back with its 32 `FiraSans-*.otf` paths and nothing from Mono or Condensed, at `4.202` from `archive/refs/tags/4.202.zip`. Condensed is the same shape at the same tag.

## 4. Stateful shards

Two packages whose download URL carries no version now use `static` with `state`, so komac derives the version during installer analysis and the response header stops an unchanged file being reanalyzed every run.

| Package | Version source | State header | Dry run |
| --- | --- | --- | --- |
| `reaConverter.reaConverterLite` | `productVersion` (inno exe) | `etag` | 8.0.234, against a manifest at 8.0.233 |
| `Ginkgo.TimeTravel` | `displayVersion` (nested `TimeTravel.msi`) | `last-modified` | 2.0.38, matching |

Neither gets a seeded `version-state` file, since I cannot tell from outside whether upstream is still on the version WinGet has, and a wrongly seeded file would suppress the first update.

No architecture overrides are set, per the CONTRIBUTING guidance not to add them preemptively. komac inferred `x86` for both, matching the manifests.

## 5. Removing Microsoft.TeamsMeetingAddin

Its three installer URLs are `go.microsoft.com` fwlinks, and `linkid=2196106` now resolves to `MSTeams-x64.msix` — the Teams client, not the meeting addin. The package no longer describes what it installs and there is nothing stable to point a shard at, so it is removed rather than covered.

## Verification

Every changed shard is dry-run by CI on each push, and I read the generated manifests rather than relying on the exit code. That mattered three times here: the VS Code `json` attempt and the Fluent icons `json` attempt both exited zero while producing wrong manifests, and the `Google.OpenSans` failure only showed up as one shard short of the total.

The three font versions in §3 were also parsed locally from the font binaries at the current head commit, using the same `name`-table reader the existing script shards use.

## Not covered

- **Pinned to a commit with no usable version source** — no upstream tags, and no derivable version: iA Writer Duo/Mono/Quattro, Quicksand, Consolas Powerline, Gohufont, SF Mono Ligaturized, Sutton SignWriting, Shafarik, SEI Sunuwar, Monoid, Meslo LG, Microsoft Scribble, .NET Native Runtime 1.4. Quicksand is close, but its font reports `3.000` where the manifest says `3`, which would create a stray version directory.
- **`Mozilla.Fira.Mono`** — same archive and tag as its siblings, but its package version is 3.206 against tag 4.202, so a tag-driven shard would submit a spurious 4.202.
- **Unversioned download URL, no safe version source**: Cinebench (2.7 GB archive, too heavy for the dry-run job), MDE Client Analyzer (+ Preview) (nested file is a `.cmd`), Kusto Explorer (already self-hosted at a versioned URL), HarmonyOS Sans, Lato, Aptos, Fluent Calibri, OneNote (its version is the Office build, not `setup.exe`).
- **Frozen or archived upstream** with no release feed: LOIC, HOIC, Banshee, NRaas Packer, Bitstream Vera, GNU FreeFont, MSXML 4, VCRedist arm32, VCLibs 12, UI.Xaml 2.0–2.4, .NET Native Framework/Runtime, IIS Express, Orca, App Verifier, MSI TrueColor, and similar.
- **Self-hosted or vendor-portal downloads** with no public version feed: the ConfigMgr tools, Deployment Toolkit, eDiscovery Export Tool, Intune Device Inventory and Endpoint Privilege Management, Global Secure Access Client, Team Explorer, AVerMedia, FlexRadio, OneLogin, Onix, Samsung NVMe, Wishingtooth, and the rest.

https://claude.ai/code/session_01J9sYCccZRbyWSEQKmA8pod


shards/json/Microsoft.VisualStudio.Code.json
```diff
@@ -0,0 +1,16 @@
+{
+	"$schema": "https://anthelion.unownplain.dev/schema.json",
+	"strategy": "page-match",
```
pl4nty, line 3: can't we use the json strategy?

shards/json/SILGlobal.Kanchenjunga.Font.json
```diff
@@ -0,0 +1,9 @@
+{
+	"$schema": "https://anthelion.unownplain.dev/schema.json",
+	"strategy": "sort-versions",
```
pl4nty: can we use https://github.com/silnrsi/font-kanchenjunga/releases instead?

shards/json/Microsoft.FluentUISystemIcons.Filled.Font.json
```diff
@@ -0,0 +1,11 @@
+{
+	"$schema": "https://anthelion.unownplain.dev/schema.json",
+	"strategy": "sort-versions",
```
pl4nty, line 3: could we use tags like https://github.com/microsoft/fluentui-system-icons/releases/tag/1.1.334

shards/json/Microsoft.VisualStudio.Code.json
pl4nty: Yes — switched to `json` in fd41091.

`update.code.visualstudio.com` serves every Windows artefact from a version-addressed path, so the build commit isn't needed at all. I checked all six against 1.130.0 and each 302s to the same CDN URL the manifest currently pins:

```
/{version}/win32-x64-user/stable      -> VSCodeUserSetup-x64-1.130.0.exe
/{version}/win32-x64/stable           -> VSCodeSetup-x64-1.130.0.exe
/{version}/win32-x64-archive/stable   -> VSCode-win32-x64-1.130.0.zip
```

...and the three arm64 equivalents. So `productVersion` alone drives everything and the shard is now six literal templates with no regex.

Your question also surfaced a bug in what I had: the page-match regex captured only `major.minor` to reproduce the existing `1.130` directory, which would have submitted a patch release like 1.129.1 as `1.129`. The `json` version can't do that since it uses `productVersion` verbatim.

One consequence worth your call: the package version becomes `1.130.0` rather than `1.130`, so the first run will add a `1.130.0` directory next to the existing `1.130`. That matches the `1.129.1` you already have (three fields when the patch is non-zero) but is a change from how `1.130` was authored. If you'd rather keep trimming the trailing `.0`, say the word and I'll take it back to `page-match` with a regex that strips `.0` only when the patch is zero — but that's the clever-regex option, and `json` is the cleaner one.

I'll confirm architecture detection from the dry-run output once CI finishes: these URLs have no file extension and contain `win32`, so if komac infers x86 from the path rather than from the resolved filename I'll add explicit `|x64`/`|arm64` overrides.

---

shards/json/SILGlobal.Kanchenjunga.Font.json
pl4nty: Yes — switched in fd41091. Tags are `v{version}` and the asset is `Kanchenjunga-{version}.zip`, both confirmed against v2.001.

I applied the same change to `SILGlobal.Kedebideri.Font` since it's the same situation (`silnrsi/font-kedebideri`, v3.001 verified) — shout if you'd rather I'd left that one alone.

Both are now `github-release` instead of scraping the software.sil.org download listing, which also moves them up your strategy preference order.

---

shards/json/Microsoft.FluentUISystemIcons.Filled.Font.json
pl4nty: Done in 395647a — both Filled and Light are now `github-release`.

`raw.githubusercontent.com` resolves a tag as a ref, so the font URL works unchanged with `{version}` in place of the commit SHA. Verified against 1.1.334 for all four styles:

```
raw.githubusercontent.com/microsoft/fluentui-system-icons/1.1.334/fonts/FluentSystemIcons-Filled.ttf     200
raw.githubusercontent.com/microsoft/fluentui-system-icons/1.1.334/fonts/FluentSystemIcons-Light.ttf      200
```

The tags carry no `v` prefix, so `{version}` is the tag verbatim. This also drops the `(1\.1\.\d+)` regex, which had the shard pinned to the 1.1.x series — it'll now follow a 1.2 or 2.0 tag if one ever lands.

There are no release assets on these tags (`/releases/download/1.1.334/FluentSystemIcons-Filled.ttf` 404s), so the raw URL is still the right source — the change is only to how the version is discovered.

Two things I did not do, both your call:

- I left `Regular` and `Resizable` on `sort-versions`. They're outside this PR and currently working, so I didn't want to touch them until this pattern is confirmed green. Happy to fold them in here or send a follow-up so all four match.
- I couldn't verify from my sandbox that this repo publishes actual Release objects rather than bare tags — github.com is 403 for me here, and `github-release` reads the releases API, not tags. The CI dry run is the real check; if it comes back empty I'll revert these two to `sort-versions` and say so.

---

pl4nty: Reverted in b0ec39a — `github-release` can't work here, and CI caught it:

```
❌ Error running Microsoft.FluentUISystemIcons.Filled.Font.json
Error: Unexpected GitHub release redirect: https://github.com/microsoft/fluentui-system-icons/releases
```

The repo publishes tags but no Release objects, so `/releases/latest` redirects to the releases index. `/releases/tag/1.1.334` renders for a bare tag, which is what made it look like a release. `github-release` reads the releases API, so it has nothing to find.

Both are back on `sort-versions` over `tags.atom`, byte-identical to the `Regular` and `Resizable` shards apart from the style name. So the four stay consistent, and the `Regular`/`Resizable` follow-up I offered is moot.

The raw-URL-at-a-tag part of the idea does hold — that's what the shard already does, it just has to discover the tag from the atom feed rather than the releases API.

---

shards/json/Microsoft.VisualStudio.Code.json
pl4nty: Have to walk this back — `json` produces a broken manifest. Reverted to `page-match` in b0ec39a.

The shard exited zero and CI counted it as a pass, so this only showed up when I read the generated manifest. komac collapsed the six `update.code.visualstudio.com` redirects into **two** installers, both arm64:

```yaml
Installers:
- Architecture: arm64
  InstallerUrl: https://update.code.visualstudio.com/1.130.0/win32-arm64/stable
  InstallerSha256: D107363D...
- Architecture: arm64
  NestedInstallerType: portable
  InstallerUrl: https://update.code.visualstudio.com/1.130.0/win32-arm64/stable
  InstallerSha256: D107363D...
```

Every x64 entry and both user-scope entries are gone. Compare the `page-match` run on 31001db, which produced all six with correct architectures, scopes and product codes. The redirects are the difference — komac needs the resolved CDN URLs, and getting those means capturing the build commit, which `json` has no way to express.

So `page-match` it is, but the version bug from my earlier regex is fixed properly rather than papered over:

```
"version":"(?<commit>[0-9a-f]{40})","productVersion":"(?<full>(?<version>\d+\.\d+(?:\.[1-9]\d*)?)(?:\.0)?)"
```

`version` is what the package gets, `full` is what goes in the installer filenames. Checked against the shapes that matter:

| productVersion | package version | filename version |
| --- | --- | --- |
| 1.130.0 | 1.130 | 1.130.0 |
| 1.129.1 | 1.129.1 | 1.129.1 |
| 1.129.10 | 1.129.10 | 1.129.10 |

A trailing `.0` is trimmed, a real patch is kept — matching both the `1.129.1` and `1.130` directories you already have, so no duplicate `1.130.0` gets created either.

If you'd rather not have the nested capture, the alternative is dropping the trim and letting the package version be `productVersion` verbatim. That's a simpler regex, but it adds a `1.130.0` alongside your existing `1.130` on the first run. Happy to switch if you prefer that trade.

---

shards/json/Microsoft.FluentUISystemIcons.Filled.Font.json
pl4nty: can we use a json strategy with a github api endpoint?

pl4nty: Good call — done in 08003e4:

```json
"strategy": "json",
"json": {
  "url": "https://api.github.com/repos/microsoft/fluentui-system-icons/tags?per_page=1",
  "path": "0.name"
}
```

The tags API gives the newest tag directly, `raw.githubusercontent.com` takes it as a ref, and the `(1\.1\.\d+)` regex is gone so the shard is no longer pinned to the 1.1.x series. I matched the `?per_page=1` style from `GarudaLinux.FireDragon`.

Waiting on CI to confirm it resolves 1.1.334 like `sort-versions` did — `api.github.com` is 403 from my sandbox so I couldn't check the response shape or ordering directly.

Two things I can't settle from here, both yours to weigh:

- `/tags` is ordered by the API, not semantically sorted. Fine today, but it wouldn't self-correct the way `sort-versions` does if the ordering ever surprises us — e.g. 1.1.99 vs 1.1.100.
- This is the only shard in the repo hitting `api.github.com`; the others use `builds.dotnet.microsoft.com`, GitLab and Codeberg. If anthelion's `json` fetch doesn't pass `GITHUB_TOKEN`, a full scheduled run shares the 60/hr unauthenticated limit. `github-release` presumably goes through the authenticated client, but the `json` strategy is a plain fetch as far as I can tell.

I left `Regular` and `Resizable` on `sort-versions` for now — happy to convert them once you're satisfied with this shape.

---

pl4nty: CI came back and the ordering risk landed immediately — reverted in 9951eec.

`fluentui-system-icons` is a monorepo, and its tags include package tags alongside the icon versions. The newest one right now is not a version at all:

```
PackageVersion: react-icons-svg-sprite-subsetting-webpack-plugin@0.0.5
InstallerUrl: https://raw.githubusercontent.com/microsoft/fluentui-system-icons/react-icons-svg-sprite-subsetting-webpack-plugin@0.0.5/fonts/FluentSystemIcons-Filled.ttf
```

That URL happens to resolve, because the tag is a real ref — so the shard exited zero and the run reported 57/57. It would have submitted `react-icons-svg-sprite-subsetting-webpack-plugin@0.0.5` as the package version.

`json` takes whatever sits at the path, with no way to filter to version tags or sort them, so it can't work against this repo regardless of endpoint. Both shards are back on `sort-versions` over `tags.atom`, identical to `Regular` and `Resizable`.

If you'd still rather go through the API, `sort-versions` can point at it instead of the atom feed — it does the filtering and sorting the `json` strategy lacks:

```json
"sortVersions": {
  "url": "https://api.github.com/repos/microsoft/fluentui-system-icons/tags?per_page=100",
  "regex": "\"name\":\"(1\\.1\\.\\d+)\""
}
```

Needs `per_page=100` to see past the package tags, and still carries the `api.github.com` rate-limit question, which is why I didn't just switch to it.

Separately, worth loosening the regex on all four shards to drop the hardcoded `1.1.`:

```
releases/tag/(\d+(?:\.\d+)+)"
```

The `releases/tag/` anchor keeps package tags out, since those don't start with a digit after the slash, while letting a future 1.2 or 2.0 through. That gets your original intent without the monorepo problem. Say the word and I'll apply it to Filled, Light, Regular and Resizable together.

---

shards/json/Microsoft.VisualStudio.Code.json
pl4nty: Removed in 1da646b — `productVersion` goes in verbatim now.

```
"version":"(?<commit>[0-9a-f]{40})","productVersion":"(?<version>\d+\.\d+\.\d+)"
```

The nested capture is gone with the trim, and `{version}` covers both the package version and the installer filenames.

One detail worth noting since it nearly bit: `version` has to be a named group here rather than a positional one. `commit` comes first and a named group still occupies capture slot 1, so an unnamed `(\d+\.\d+\.\d+)` would have been group 2 and the commit SHA would have been taken as the version. Caught it testing the regex before pushing.

Heads up on what this does on the first run: the package version becomes `1.130.0`, so it lands alongside your existing `1.130` rather than updating it.

---

pl4nty: Found it — #549 landed on main after I branched, and I was checking a stale `origin/main`. My apologies for the previous reply.

Merged main and renamed the shard in 974189f:

- `shards/json/Microsoft.VisualStudio.Code.json` → `shards/json/Microsoft.VisualStudioCode.json`

The filename has to equal the package identifier or the shard is simply never found, so it would have silently stopped matching once #549 merged.

That also resolves the caveat from my last reply. #549 renumbered the directory to `1.130.0`, so dropping the trim isn't just simpler — it's now the only correct behaviour. The shard resolves `1.130.0`, which is the version already in the repo, so no stray directory gets created.

Verified after merging: no orphaned shards, coverage holds at 236/351, `manifests:check` and `test:manifests` pass. Waiting on the dry run to confirm the shard is picked up under the new name.

---

shards/script/Google.Fira.SansExtraCondensed.Font.ts
```diff
@@ -0,0 +1,50 @@
+import { defineShard } from 'anthelion';
+import { githubClient } from 'anthelion/github';
+import ky from 'ky';
+
+// The upstream repo has no versioned releases, so track the default branch and
+// pin URLs to its head commit; the version comes from the font's name table.
+function fontVersion(bytes: Uint8Array): string {
```
pl4nty: this function should be shared

pl4nty: Done in 00fd5de — it's `scripts/font-version.ts` now, imported via the `@/` alias. It was copied byte-for-byte into seven shards (Tinos, SpaceMono, OpenSans and EgyptianOpenType already had it), so that's 182 lines out.

It can't live under `shards/`: the Test workflow basenames every changed file matching `shards/**` into a package ID, so a module there gets run as a shard.

Checked the alias resolves at runtime under bun, not just for typecheck.

---

shards/script/Google.Fira.SansExtraCondensed.Font.ts
```diff
@@ -0,0 +1,26 @@
+import { defineShard } from 'anthelion';
+import { githubClient } from 'anthelion/github';
+import ky from 'ky';
+
+import { fontVersion } from '@/scripts/font-version';
+
+// The upstream repo has no versioned releases, so track the default branch and
+// pin URLs to its head commit; the version comes from the font's name table.
+export default defineShard(async () => {
```
pl4nty: a lot of this logic might fit in fontVersion too

shards/script/Google.Fira.SansExtraCondensed.Font.ts
pl4nty: Agreed — done in fb38af1. `branchFontShard` now owns the `listCommits`, head sha, font fetch and return block, so this shard is just:

```ts
export default defineShard(() =>
	branchFontShard({
		owner: 'googlefonts',
		repo: 'FiraGFVersion',
		path: 'fonts/FiraSansExtraCondensed-Regular.ttf',
	}),
);
```

`urls` defaults to the repo archive, which is what five of the seven use. EgyptianOpenType and OpenSans install the single font they read, so they pass their own. VictorMono takes its version from `package.json` rather than a name table, so it keeps its body and shares `headCommit` only.

Checked the new config resolves to the same read and installer URLs as before, and ran `branchFontShard` against a stubbed client — 4.203, the archive URL, head sha as state.

---

PR #577, opened 2026-07-26
New package: Mozilla.Firefox.Portable version 153

DandelionSprout:
Checklist for Pull Requests

  - Relates to https://github.com/microsoft/winget-pkgs/pull/400642.

Manifests

- [x] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [x] Have you tested your manifest locally with `winget install --manifest <path>`?

---


DandelionSprout:
Oh, right.

manifests/m/Mozilla/Firefox/Portable/153/Mozilla.Firefox.Portable.installer.yaml
```diff
@@ -0,0 +1,51 @@
...
+- html
+- pdf
+- shtml
+- svg
+- webp
+- xht
+- xhtml
+Installers:
+- Architecture: x64
+  InstallerUrl: https://download-installer.cdn.mozilla.net/pub/firefox/releases/153.0/win64/multi/Firefox%20Setup%20153.0.msix
```
pl4nty, line 36: An MSIX is not necessarily portable, is there a proper portable build of Firefox available instead?

DandelionSprout: It's pretty complicated. Because I know of at least 3 different types of officially Mozilla-published and -hosted portable Firefox versions, yet the Firefox devs refuse to acknowledge any of them even when they are shown their existence and officiality up-close (I have first-hand experience with those devs during some chaos over at pkgs last spring).

#### Approach 1, used by this PR:

Take the stable MSIX version, force Winget (The CLI tool) to treat it as a ZIP (which is what an MSIX/AppX essentially is), confirm the main .exe can be launched from an extracted MSIX (Around 50% of cases do, and Firefox is one of them), submit it as a portable.

I learned the trick some months ago to remarkable success over at pkgs, where it is currently in use for e.g. KDE Okular and at least 15 other packages.

#### Approach 2, more controversial than anyone would think:

Relatively deeply hid on Mozilla's hosting service is the "candidates" channel (https://ftp.mozilla.org/pub/firefox/candidates/), presumably in the meaning of "Release Candidate". It seems to be the midway point between stable and Beta, and are published 3〜4 days before the true stable version goes live.

And those Candidates builds do in fact have fully portable ZIPs, which means that `https://ftp.mozilla.org/pub/firefox/candidates/153.0-candidates/build1/win64/en-US/firefox-153.0.zip` would've worked too as an InstallerUrl.

The problem with that, is that Mozilla Corporation straight up denies the Candidates channel's existence (or at least that it's them who actually make/build/push/host it), Tiananmen-style. Had we used that channel, they'd go far to discredit the package.

#### Approach 3, very impractical for many reasons:

Slightly more publicly acknowledged by them is the "Unbranded" channel, as explained on https://wiki.mozilla.org/Add-ons/Extension_Signing#Unbranded_Builds; the InstallerUrl-s in that case would be along the lines of https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/CapnPdAbQemn76V4BJVfeA/runs/0/artifacts/public/build/target.zip. Though it's seemingly closest to Firefox Beta and/or ESR, the Firefox devs treat it as closer to a pre-alpha. And the fact it uses a rotating hash would make this approach very impractical, as is the fact that the "Unbranded" channel does not use the Firefox brand, which would confuse many users.

pl4nty: Sorry about the wait. I think your approach is the best of the three, but I can't see why it's necessary instead of just installing the MSIX directly

DandelionSprout: The market for a portable Firefox is big enough that a 3rd-party fork for portability got a Wikipedia article for it: https://en.wikipedia.org/wiki/Firefox_Portable

And in contrast to the one on Wikipedia, this PR's portable Firefox is 1st-party.

pl4nty: Hmm, that seems fair. I'd just prefer that you use Mozilla.Firefox as the identifier, with version 153.0.3 to match the upstream format. And include msix/nullsoft/portable in the one manifest

DandelionSprout: That could take me quite a while.

DandelionSprout: I ended up doing it, but Unown's analysis over at https://github.com/pl4nty/winget-extras/pull/868#issuecomment-5297643667 means that I don't think a multi-installer-type manifest is a good idea.

Even more so since this PR's portable version is *de jure* locale-neutral, which is infamously not the case for Firefox's Nullsoft installers.

PR #587, opened 2026-07-27
Remove version: Microsoft.VisualStudioCode version 1.130.0

pl4nty:


pl4nty:
On second thought, maybe we keep this just in case vscode maintainers go awol again

PR #593, opened 2026-07-27
chore: add more shards

pl4nty:
Adds shards for the six packages flagged by `repository/shard-coverage`. Three have a usable public version source; the other three get a `.disabled` shard with the reason.

Checklist for Pull Requests

  - N/A, shards only

Manifests

- [ ] Have you checked that there aren't other open [pull requests](https://github.com/pl4nty/winget-extras/pulls) for the same manifest update/change?
- [ ] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)?

No manifests are changed by this PR.

---

## Shards added

**`FlexRadioSystems.SmartSDR`** — `sort-versions` over https://www.flexradio.com/software/, which lists every release as `SmartSDR vX.Y.Z`; greatest wins (currently 4.2.20, vs 3.4.24 in the repo). Installers live at a predictable CDN path, and v4 ships x64 and arm64 MSIs, both of which return 200.

**`AVerMedia.RECentral`** — `page-match` on the support download listing. The installer filename carries a build stamp that can't be derived from the version (`RECentral_4_v4.7.114.1_24121801_x64.exe`), so the regex takes a named `build` capture and the URL template reuses it. Anchored to `/web_release_www/software/` because the same listing carries old device-bundled RECentral 4 builds under per-model directories. RECentral 4 is on AVerMedia's end-of-service list, so this will likely only ever fire once (4.7.44.1 → 4.7.114.1).

**`DitDitDit.GeminiAmplifier`** — `page-match` on the "The current kit is version 1.2.0, February 22nd 2022 [Link]" line. The download is a Dropbox share whose slug is random per file, so the regex captures the whole URL from the link that follows the version and the template just swaps `?dl=0` for `?dl=1`.

## Shards disabled

**`AVerMedia.AssistCentral`** — the non-Pro Assist Central is gone from AVerMedia's download catalog; the full listing at https://www.avermedia.com/support/download only carries Assist Central Pro (4.0.x, a different product line from this package's 1.2.x).

**`MSI.TrueColor`** — `www.msi.com` answers non-browser requests with an Akamai 403, including its `/api/v1/product/support` endpoint, and True Color is only published per notebook model with an opaque hash in the filename (`Install%20TrueColor20260616exe_5.1.24.0_0xf40c4328.zip`).

**`OneLogin.DesktopPro`** — the `onelogin-desktop-release` S3 bucket denies listing and any key other than a known-exact one, and https://support.onelogin.com/onelogin/download-new-releases redirects to `/Login`.

## Testing

`bun test:shard --dry-run` could not be run here: `bun install` fails in this environment because the `anthelion` dependency is fetched from the GitHub tarball API, which the sandbox blocks. Instead, each regex was run against the live page HTML (correct version and captures, no competing matches), every resulting installer URL was confirmed to return 200, and the three JSON shards were validated against https://anthelion.unownplain.dev/schema.json. A dry run before merge would still be worthwhile.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KysDnNUaM9nc38DmDQPZtQ)_


pl4nty:
CI is green — `bun test:shard --dry-run` resolves 3/3 shards: Gemini Amplifier 1.2.0 (hash matches the manifest already in the repo), RECentral 4.7.114.1, SmartSDR 4.2.20.

One thing worth knowing before merge: the SmartSDR dry run generates an **x64-only** installer manifest even though the shard lists both MSIs. Both files exist and download fine, and the arm64 one is correctly stamped `Arm64;1033` in its MSI summary info, but komac collapses it into the x64 entry — adding `|arm64` didn't change the output. Most likely the two builds share a ProductCode (`{A6189D80-D83F-471F-B158-37AC6056D189}`), so komac treats them as one installer.

I left the arm64 URL in the shard: it's a correct upstream URL and the output would pick it up for free if komac's handling changes. The cost is one extra ~650&nbsp;MB download per release. Happy to drop it to x64-only if you'd rather the shard match what actually lands in the manifest.

Also note the installer-object form for architecture overrides in the published schema isn't accepted by the pinned anthelion (`fc5452e`), whose Zod validator only takes strings — the `url|arch` suffix is the working syntax here.

---

PR #628, opened 2026-07-27
chore(linter): use anthelion-komac's YAML parser

UnownPlain:
Replaces Bun.YAML and yaml with anthelion-komac's YAML parser so we don't need 2 different parsers. Anthelion komac's implementation is just a pretty much just a wrapper around the https://github.com/bourumir-wyngs/serde-saphyr and https://github.com/bourumir-wyngs/granit-parser libraries already used to generate and parse manifests so YAML conformance should be identical serde-saphyr.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -15,7 +15,6 @@
         "oxfmt": "0.60.0",
         "oxlint": "1.75.0",
         "oxlint-tsgolint": "7.0.2001",
-        "yaml": "2.9.0",
       },
     },
   },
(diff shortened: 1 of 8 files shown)


pl4nty:
thanks, hopefully this'll avoid any more parser bugs. would you accept PRs to anthelion-komac? a lot of your changes are pretty useful, and I have many fixes/analysers that would be slow to merge upstream

UnownPlain:
Sure, as long as you're willing to maintain the analyzers you're adding.

PR #629, opened 2026-07-28
ci: trigger automerge on PR open

UnownPlain:
GitHub search sometimes likes to break so this should be more reliable.


pl4nty:
thanks! I started with this and merged into one workflow, thinking it'd be cleaner. but didn't consider reliability

PR #639, opened 2026-07-30
fix: linter improvements

UnownPlain:
I would've tried to stack these changes since GitHub has finally released stacked diffs a few hours ago, but they don't support cross fork stacks :(

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,9 +7,9 @@
 		"fmt": "bun --bun oxfmt",
 		"lint": "bun --bun oxlint --type-check --type-aware",
 		"test:manifests": "bun test scripts",
-		"manifests:generate": "bun scripts/manifest-linter/generate-manifest-types.ts",
-		"manifests:check": "bun scripts/manifest-linter",
-		"manifests:fix": "bun scripts/manifest-linter --fix"
+		"manifests:generate": "bun --experimental-http2-fetch scripts/manifest-linter/generate-manifest-types.ts",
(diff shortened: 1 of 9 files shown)


pl4nty:
thanks! I left feedback about fork support during the stacked diffs preview, but I guess their deadline was end of July. I'm happy to give you write access here if you want?

UnownPlain:
Thanks, having write access would be great. It looks like they haven't shipped cross fork support due to security concerns.

From https://news.ycombinator.com/item?id=49114656:
> Yes it will be coming! The reason it's taking a bit longer is because of the automated rebase that happens after you merge part of a stack. There are some legitimate security concerns because of this so multi-fork stacks (a stack which includes multiple different forks) are probably out of the question for now.
> 
> We will support a stack that is fully contained within a single fork, where the entire stack targets the original repo.
> 
> For example, a contributor who has a fork (user/buzz) of the original repo (org/buzz) could create the following stack:
> 
> ``` frontend → PR #3 (base: user/buzz:api-endpoints) api-endpoints → PR #2 (base: user/buzz:auth-layer) auth-layer → PR #1 (base: org/buzz:main) org/buzz:main (trunk) ```

Hopefully they release it soon since it would probably be useful for upstreaming some of Anthelion komac changes.

pl4nty:
Sent the invite. I was hoping to use stacks with komac too, rebasing and cherry-picking feature branches has been a headache

PR #652, opened 2026-08-01
fix: linter warnings, remove GameJolt, automerge Renovate PRs

pl4nty:
Clears the manifest linter's outstanding `repository/shard-coverage` warnings, removes a package that cannot be kept installable, and extends the automerge workflow to Renovate.

## Checklist for Pull Requests

- Relates to: no winget-pkgs issue — repository tooling only.

## Manifests

`GameJolt.GameJolt` is removed (see below). No manifests are added or modified. `bun manifests:check` reports `Manifests OK (1395 files)` with no `shard-coverage` warnings (previously 3 warnings over 1398 files).

## Removed: GameJolt.GameJolt

Game Jolt serves the client installer only from `download.gamejolt.net` behind a signed URL. The client is published as a game on Game Jolt itself (game `362412`, build `1621836`), so the version and a download URL are both retrievable:

```
GET  /site-api/web/discover/games/packages/362412
     -> releases[0].version_number == "1.39.4"
POST /site-api/web/discover/games/builds/get-download-url/1621836
     -> { payload: { url: "https://download.gamejolt.net/<sig>,<exp>,7/..." } }
```

Fetching a freshly minted URL returns sha256 `c411ab8f…ff5a496`, matching the installer manifest byte for byte — so this is a live, correct package, not a dead one.

The problem is lifetime. `<exp>` is a unix timestamp exactly 24h after the URL is minted, and Game Jolt exposes no server-side redirect to the signed URL (`/get/build` and `/downloads/build` both return SPA HTML that mints it client-side). Any URL committed to a manifest is dead within a day.

That has already happened. The URL shipped in `GameJolt.GameJolt.installer.yaml` carried `,1785248918,`, which expired **2026-07-28** and now returns HTTP 400 — `winget install GameJolt.GameJolt` fails today. Refreshing it would only restart the same 24h clock, so the manifests and the shard are removed rather than carried in a permanently broken state.

This is the first package removal in the repository, so there was no existing convention to follow; the manifest directory and the shard file are simply deleted.

## Shard coverage

**`Microsoft.AccessibilityInsights`** gets a real shard. Releases are tagged `v1.1.2924.01` while the package is versioned `1.1.2924.1`, so the download is templated from `{github.rawTag}` and the padding zero is dropped from the resolved version. The `Test / Test` dry run confirms it regenerates `PackageVersion: 1.1.2924.1`, matching the manifest already in the repository.

**`Microsoft.WindowsSDK.BuildTools`** is recorded as a disabled shard. The only publisher is nuget.org, and `…/api/v2/package/…` redirects to a `.nupkg`. Komac rejects that extension during installer analysis (verified locally: identical bytes served as `.zip` analyse fine, as `.nupkg` they fail), so a shard cannot re-derive the nested `bin\{sdk-build}\{arch}\` tool paths that change with every release. Disabled shards are the convention already used for the ~55 other packages that cannot be updated automatically.

## Automerge

`Automerge` now fires for `renovate[bot]` as well as `winget-extras-bot[bot]`. It still shells out to `gh pr merge --squash --auto`, so GitHub holds the merge until the required checks pass — Renovate PRs that fail CI stay open. Renovate opens its branches in this repository, so the existing same-repo guard continues to apply.

## Not addressed

`repository/upstream-versions` still warns that `Microsoft.MSXML.4` `4.0-SP3` exists in winget-pkgs. Resolving that means removing the package from this repository rather than changing tooling, so it is left for a maintainer to decide.

https://claude.ai/code/session_01JHGTkokVb9K32E5yrqkcku


DandelionSprout:
* *"(…) a unix timestamp exactly 24h after the URL is minted, and Game Jolt exposes no server-side redirect to the signed URL (…)"*

That is admittedly fully correct. There is however some context that at least I personally considered to be mildening factors:
* Other apps I ran tests on earlier this week to see when they expired, would expire after barely even 5 minutes. Realtek's links would expire after 1 minute. 24 hours felt like an eternity in comparison at the time.
* There is sort-of precedence for it in pkgs, in which `Microsoft.SafetyScanner` gets updated every 5 hours or so. However, `Microsoft.SafetyScanner` links would expire after 10 days (or at least that's what Microsoft has claimed).

pl4nty:
@DandelionSprout I didn't want to introduce daily churn into this repo. But I think I can readd it by hosting the installer on my own storage

DandelionSprout:
Understandable.

PR #656, opened 2026-08-01
chore: enable automerge in Renovate config

UnownPlain:
Renovate has native support for automerge so we don't need to configure a workflow for it.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -3,6 +3,10 @@
 	"extends": ["github>pl4nty/.github"],
 	"packageRules": [
 		{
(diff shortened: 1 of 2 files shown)


pl4nty (approved):
I thought Renovate's automerge required a rerun, so it'd be slower, and figured we already had an automerge workflow. But in hindsight, only automerging patch or lower is probably better

UnownPlain:
> I thought Renovate's automerge required a rerun, so it'd be slower

Not sure what you mean by this. In my experience it's enabled instantly after the PR is opened.

PR #790, opened 2026-08-09
ci(publish): add sourceversion object metadata

UnownPlain:
I haven't tested this, but this should theoretically allow us to set the `x-ms-meta-sourceversion` header by setting up a response header transform rule like the following:

```
http.host eq "winget.tplant.com.au" and http.request.uri.path in {"/cache/source.msix" "/cache/source2.msix"} and http.response.code eq 200 and has_key(http.response.headers, "x-amz-meta-sourceversion")
```

```
Operation:   Set dynamic
Header name: x-ms-meta-sourceversion
Expression:  http.response.headers["x-amz-meta-sourceversion"][0]
```

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -94,8 +94,14 @@ jobs:
           - name: Setup and run IndexCreationTool
             working-directory: index
             run: |
-              $version = Get-Date -Format yyyy.M.d.hm
+              $now = [DateTime]::UtcNow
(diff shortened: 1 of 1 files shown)


pl4nty:
No luck unfortunately, I even set up some local creds to test. rclone doesn't set the metadata, but even if I set it with the AWS CLI, it's not returned in headers. [Looks like an undocumented limitation](https://www.answeroverflow.com/m/1441237622591852638)

PR #852, opened 2026-08-13
feat(linter): add ignore config

UnownPlain:
Adds a config file to ignore linter rules for specified manifests. Also adds back `--deny-warnings` to CI since there's a way to ignore them now.


pl4nty (approved):
thanks, this is much cleaner. I've been wondering about a similar structure for validation waivers, what do you think?

UnownPlain:
> I've been wondering about a similar structure for validation waivers, what do you think?

Sounds good. I was thinking of that too since we have some interactive only packages which can't pass validation.

PR #872, opened 2026-08-14
feat: add package move script

UnownPlain:
Adds a small script to make it easier to move packages to a new identifier.


pl4nty (approved):
neat, thanks

PR #902, opened 2026-08-15
New package: SichboPVR.SichboPVR.3 version 3.1.105

DandelionSprout:
Checklist for Pull Requests

- [x] Is there a related issue or pull request in [winget-pkgs](https://github.com/microsoft/winget-pkgs)? If so, add the link(s) below.
   
  - Relates to https://github.com/microsoft/winget-pkgs/pull/414884.

Manifests

- [x] Have you checked that there aren't other open [pull requests](https://github.com/pl4nty/winget-extras/pulls) for the same manifest update/change?
- [x] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [x] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)?

Note: `<path>` is the directory's name containing the manifest you're submitting.

---

PR #913, opened 2026-08-17
feat(linter): accept localized values of Proprietary and Freeware

UnownPlain:
This adds translations for Proprietary and Freeware to the linter for the [50 most common languages](https://en.wikipedia.org/wiki/Languages_used_on_the_Internet) used on in the internet. Most the of translations were pulled from the [Wikidata](https://www.wikidata.org/wiki/Q218616) [pages](https://www.wikidata.org/wiki/Q178285) and the rest were from Google Translate with some minor changes by Codex.

I think this is the best compromise for https://github.com/pl4nty/winget-extras/pull/902#issuecomment-5306324398.


pl4nty (approved):
Thanks. This seems like a good compromise, although I still feel `License` could be made machine-readable-only (rather than human-readable) in the future

UnownPlain:
We can enforce that all packages have an English manifest if that's a problem.

PR #48, opened 2026-06-16
New package: Insecure.Npcap version 1.88

pl4nty:
### Pull request has been created with [komac](https://github.com/russellbanks/Komac) v2.16.0 🚀

- Relates to microsoft/winget-pkgs#361415

PR #1017, opened 2026-08-21
New package: GyulyVGC.sniffnet version 1.5.1

pl4nty:
Checklist for Pull Requests

  - Relates to microsoft/winget-pkgs#361406

Manifests

Note: authored in a Linux environment, so `winget validate`/`winget install` weren't run locally — relying on this repo's manifest linter (`bun run manifests:check` passes) and CI sandbox validation instead.

---

Adds Sniffnet 1.5.1, a network traffic monitor, per the winget-pkgs package request above. It can't be added upstream because it requires the Npcap driver, whose winget package is stuck on an outdated version — so this manifest declares a dependency on this source's own `Insecure.Npcap` package instead.

- All three Windows MSIs (x64, x86, arm64) with SHA256 hashes computed from the downloaded installers, and ProductCode/UpgradeCode extracted from the MSI property tables (`ALLUSERS=1`, so `Scope: machine`)
- The MSI doesn't bundle Npcap; it only shows an exit-dialog reminder, so silent install works
- `github-release` shard for automated updates

https://claude.ai/code/session_0164i3Sd7WHX7yDC7oESW3FE

---
_Generated by [Claude Code](https://claude.ai/code/session_0164i3Sd7WHX7yDC7oESW3FE)_

### DIFF
--- a/manifests/g/GyulyVGC/sniffnet/1.5.1/GyulyVGC.sniffnet.installer.yaml
+++ b/manifests/g/GyulyVGC/sniffnet/1.5.1/GyulyVGC.sniffnet.installer.yaml
@@ -1,0 +1,45 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.28.0.schema.json
+
+PackageIdentifier: GyulyVGC.sniffnet
+PackageVersion: 1.5.1
+InstallerLocale: en-US
+InstallerType: wix
+Scope: machine
+UpgradeBehavior: install
+Dependencies:
+  PackageDependencies:
+  - PackageIdentifier: Insecure.Npcap
+ReleaseDate: 2026-07-22
+Installers:
+- Architecture: x64
+  InstallerUrl: https://github.com/GyulyVGC/sniffnet/releases/download/v1.5.1/Sniffnet_Windows_x64.msi
+  InstallerSha256: 3F4DACF273B0E5270D4055A6561C5B0E1E0747EF6530F984CC42389EBC401700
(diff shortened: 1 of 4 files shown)


pl4nty:
The three Installation Validation failures are inherent to the Npcap dependency, not a manifest bug, so I'm not pushing a "fix" that would drop it:

- The Sniffnet MSIs install silently fine on their own (no bundled Npcap, `ALLUSERS=1`, exit-dialog reminder only).
- Validation runs `winget install --silent`, which resolves the `Insecure.Npcap` dependency from this source and runs `npcap-1.88.exe /S` ([log](https://github.com/pl4nty/winget-extras/actions/runs/32534814996/job/96933699513)). Free Npcap doesn't support silent installation (OEM-only), so the installer sits there until the 5-minute timeout kills the job — same on all three architectures, and it'll reproduce on every run.

Options, in order of my preference:

1. **Merge despite red validation.** The dependency chain is the point of hosting Sniffnet here (per the discussion in #48 / microsoft/winget-pkgs#361406) — installing via `winget install GyulyVGC.sniffnet --interactive` walks the user through the Npcap wizard, then installs Sniffnet.
2. **Teach validation about interactive-only dependencies**, e.g. retry with `--skip-dependencies` when the silent install times out on a dependency. Happy to do this in a follow-up PR if preferred — I didn't want to widen this one.
3. **Drop the dependency** and rely on the description/InstallationNotes telling users to install `Insecure.Npcap` first. Worst UX of the three.

Also worth noting from the #48 thread: GyulyVGC offered to open a Sniffnet PR themselves — this one may want coordinating with them.

---

PR #1054, opened 2026-08-24
feat(validation): expect timeout if installer is interactive

UnownPlain:
We can use the manifest's `InstallModes` field to check if an installer is interactive and expect a timeout. This will allow for interactive-only installers to pass validation and allow the auto-update PR to be automatically merged.


pl4nty:
Thanks, sounds good. Will just need to be careful when reviewing new packages

PR #1087, opened 2026-08-28
New package: worproject.Imager version 2.3.1

pl4nty:
Checklist for Pull Requests

  - Relates to https://github.com/microsoft/winget-pkgs/pull/422237.

Manifests

Note: `<path>` is `manifests/w/worproject/Imager/2.3.1/`.

---

This adds `worproject.Imager` version 2.3.1 (Windows on Raspberry imager), relating to #1068, which @DandelionSprout opened but couldn't add a shard for, noting:

> Reason for not having been able to make a shard: The InstallerUrl contains tag numbering that differs between versions. For this version, it's `13%2F02%2F2024`.

That's correct — the `worproject/dldserv-mirror` release tag is the date the mirror was uploaded, not a package version, and isn't derivable from the version number. On top of that, a single release bundles assets for several worproject downloads (the imager, the boot mounter, etc.), so a plain `github-release` JSON shard can't select the right file either. I added a script shard (`shards/script/worproject.Imager.ts`) instead: it fetches the latest `dldserv-mirror` release, filters the asset list down to the one matching `WoR_Release_*.zip`, and reads the version back out of that filename.

The manifest content matches what was proposed in #1068 — same version, same installer URL/hash — I re-downloaded the installer myself and recomputed its SHA256 rather than trusting the value already in that PR, and it matched. I also checked worproject's own changelog and downloads page, which both still list 2.3.1 as current.

I don't have a Windows environment here, so I couldn't run `winget validate`/`winget install` against the manifest directly. Instead I verified locally with this repo's own tooling:

```sh
bun manifests:check --deny-warnings
bun test:manifests --deny-warnings
bun fmt --check
```

All three passed. (The real `anthelion` npm dependency can't be installed in this sandbox since it needs GitHub API access to a different org than this token has, so I temporarily swapped it for pinned versions of its own dependencies to run the checks, then restored `package.json`/`bun.lock` before committing — only the manifest and shard files are included in this PR.)

This is not a replacement for #1068 — that PR is DandelionSprout's and I haven't touched it. This is a separate submission that happens to add the shard piece that was missing.

---

### DIFF
--- a/manifests/w/worproject/Imager/2.3.1/worproject.Imager.installer.yaml
+++ b/manifests/w/worproject/Imager/2.3.1/worproject.Imager.installer.yaml
@@ -11,7 +11,7 @@ NestedInstallerFiles:
 ArchiveBinariesDependOnPath: true
 Installers:
 - Architecture: x86
-  InstallerUrl: https://github.com/worproject/dldserv-mirror/releases/download/130.000000020.0000002024/WoR_Release_2.3.1.zip
+  InstallerUrl: https://github.com/worproject/dldserv-mirror/releases/download/13%2F02%2F2024/WoR_Release_2.3.1.zip
   InstallerSha256: FC3F6A98E1744ADAA8C514514C35176805AAC8BB44E16514C2A979683FA0AE1F
 ManifestType: installer
 ManifestVersion: 1.12.0
--- a/manifests/w/worproject/Imager/2.3.1/worproject.Imager.installer.yaml
+++ b/manifests/w/worproject/Imager/2.3.1/worproject.Imager.installer.yaml
@@ -1,0 +1,17 @@
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: worproject.Imager
+PackageVersion: 2.3.1
+ReleaseDate: 2024-02-13
+InstallerType: zip
+NestedInstallerType: portable
+NestedInstallerFiles:
+- RelativeFilePath: WoR.exe
+  PortableCommandAlias: wor
+ArchiveBinariesDependOnPath: true
+Installers:
+- Architecture: x86
+  InstallerUrl: https://github.com/worproject/dldserv-mirror/releases/download/130.000000020.0000002024/WoR_Release_2.3.1.zip
+  InstallerSha256: FC3F6A98E1744ADAA8C514514C35176805AAC8BB44E16514C2A979683FA0AE1F
+ManifestType: installer
+ManifestVersion: 1.12.0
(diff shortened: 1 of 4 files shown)


pl4nty:
CI's `Validate / Installation Validation (zip, x86, ...)` fails on the actual `winget install`, not on the manifest content or the shard:

```
<E> [CORE] DeliveryOptimization error: 0x80190194, extended error: 0x00000000
<E> [FAIL] ...DODownloader.cpp(361)... Exception(1) tid(830) 80190194 Not found (404).
<E> [FAIL] ...Downloader.cpp(210)... Exception(2) tid(830) 80190194 Not found (404).
```

`0x80190194`'s low word (`0x194` = 404) is Delivery Optimization reporting an HTTP 404 for the `InstallerUrl`:
`https://github.com/worproject/dldserv-mirror/releases/download/13%2F02%2F2024/WoR_Release_2.3.1.zip`

That URL is correct — it's byte-identical to the one in #1068, and I confirmed it resolves fine (`curl -L` → 200, redirecting through GitHub's normal signed-asset flow) from this Linux sandbox. But Windows' downloader 404s on it. I also found GitHub's own `releases/latest/download/<file>` convenience redirect decodes the same tag's `%2F` into literal `/`, producing a broken 3-segment path (`.../download/13/02/2024/WoR_Release_2.3.1.zip`) in its `Location` header — independent evidence that a literal `/` in a release tag is fragile across HTTP stacks, not just an artifact of my testing. My best explanation is that Windows' Delivery Optimization/WinINet path canonicalizes `%2F` to `/` before issuing the request, splitting the tag into extra path segments GitHub then 404s on — but I don't have a Windows environment to confirm the exact request DO makes or to safely try alternate encodings (e.g. double-encoding) without guessing blind.

Since the tag itself (`dldserv-mirror`'s upload date `13/02/2024`) is fixed by upstream and there's no slash-free way to reach this specific dated release's asset, I don't have a fix I'm confident in without Windows access to verify it. Flagging this rather than pushing speculative encoding changes. Note #1068 (the PR this replicates) never checked its own "tested with `winget install`" box either, so this may be a pre-existing, previously-unverified problem with that manifest content, not something introduced here.

---

pl4nty:
Correction to my previous comment: the Windows Delivery Optimization theory was wrong. The real cause was simpler — the `InstallerUrl` in the first commit was corrupted at authoring time (a shell `printf` interpreted `%2F` as its own format specifier, mangling the tag into `130.000000020.0000002024`), not a genuine encoding incompatibility. That's now fixed in the second commit, using a URL taken directly from the file rather than shell string interpolation, and CI is fully green — `winget install` succeeds against the corrected URL. Apologies for the noise.

---

DandelionSprout:
* *"Apologies for the noise."*

Apology accepted, as I've seen much wilder comment noise than this over the years.

* *"That's correct — the worproject/dldserv-mirror release tag is the date the mirror was uploaded, not a package version, and isn't derivable from the version number. On top of that, a single release bundles assets for several worproject downloads (the imager, the boot mounter, etc.), so a plain github-release JSON shard can't select the right file either."*

Good reasoning.

PR #1089, opened 2026-08-28
New package: Microsoft.SystemCenter.ServiceManager.Backup version 10.25.1523

pl4nty:
Checklist for Pull Requests

  - Relates to https://github.com/microsoft/winget-pkgs/discussions/333858#discussioncomment-18027614
  - Relates to https://github.com/microsoft/winget-pkgs/issues/329209#issuecomment-4557209250

Manifests

Note: `<path>` is `manifests/m/Microsoft/SystemCenter/ServiceManager/Backup/10.25.1523/`.

---

### About this PR

This replicates #910 ("New package: Microsoft.SystemCenter.ServiceManager.Backup version 10.25.1523"), originally opened by @DandelionSprout, as a separate PR that adds a working automated-update shard. #910 itself is untouched — this is a new, independent submission.

**Package**: Encryption Key Backup or Restore Wizard, from Microsoft System Center Service Manager 2025 — an `x86` zip archive with a portable `SecureStorageBackup.exe` nested inside.

**Why a shard turned out to be feasible**: DandelionSprout's PR body correctly notes the `InstallerUrl` (`https://download.microsoft.com/download/6314b969-ddb8-42f2-aec0-5c96e8c4630a/SCSM_2025.zip`) is non-GitHub and has no version pattern in it — it's a static, rolling-release URL. However, this repo already has a solved shard for the sibling package `Microsoft.SystemCenter.Orchestrator.RunbookDesigner`, which downloads from the *same* `download.microsoft.com` GUID folder (`SCO_2025.zip` instead of `SCSM_2025.zip`) and has the identical problem. Its shard uses Anthelion's `static` strategy with `"version": { "source": "product" }`, which tells komac to read the version from the installer's own file/product metadata rather than the URL.

I verified this works for this package too: I downloaded `SCSM_2025.zip` myself, extracted `Tools/SecureStorageBackup/SecureStorageBackup.exe`, and read its Windows version resource with `pefile` — `ProductVersion` is `10.25.1523.0`, matching the proposed `PackageVersion: 10.25.1523` exactly. I also verified the SHA256 of my own download against the hash quoted in #910 — they match byte-for-byte (`3001D0F1...D41B1`).

**Shard**: `shards/json/Microsoft.SystemCenter.ServiceManager.Backup.json`, `static` strategy, `version.source: product`, `state` tracked via the `etag` response header on the same static URL (so re-analysis is skipped when the upstream file hasn't changed), `replace: true` since it's a rolling-release URL like its sibling.

**Manifest differences from #910's proposal** (kept for correctness/consistency with repo conventions):
- Schema pinned to `1.12.0` (majority convention in this repo; `1.28.0` also appears but is less common) instead of `1.28.0`.
- Dropped the `FileExtensions: [bin]` field — not present on any comparable portable/zip manifest in this repo and not something I could substantiate.
- Fixed a typo in the proposed tags (`system-center-storage-manager-backup` → `system-center-service-manager-backup`, matching the actual package name "Service Manager").
- Locale file uses `en-US` (this repo's dominant convention for Microsoft manifests) rather than the bare `en` in #910; dropped the additional `nb` (Norwegian) locale for simplicity — happy to add it back if wanted.

### Validation

No Windows environment is available in this sandbox, so I could not run `winget validate`/`winget install` directly. Instead I verified with this repo's own tooling (the real `anthelion` npm dependency needs GitHub API access unavailable here, so per this repo's testing workaround I temporarily swapped it for its pinned dependency graph, ran the checks, then restored `package.json`/`bun.lock` exactly — neither is part of this PR's diff):

- `bun manifests:check --deny-warnings` → `Manifests OK (2211 files)`
- `bun test:manifests --deny-warnings` → `89 pass, 0 fail`
- `bun fmt --check` → clean (only the temporarily-swapped `package.json` flagged, which was reverted before committing)

I could not run `bun test:shard Microsoft.SystemCenter.ServiceManager.Backup --dry-run` (the `anthelion-test` binary isn't available without the real `anthelion` package in this sandbox), so the shard's live behavior against Anthelion's actual pipeline is unverified beyond the manual file-metadata check above and structural parity with the working `RunbookDesigner` shard.

---


DandelionSprout:
* *"it's a static, rolling-release URL. However, this repo already has a solved shard for the sibling package `Microsoft.SystemCenter.Orchestrator.RunbookDesigner`, which downloads from the *same* download.microsoft.com GUID folder (SCO_2025.zip instead of SCSM_2025.zip) and has the identical problem."*

Oh, right. I was however previously personally unsure whether Service Manager and Orchestrator could be considered the same app suite, but Claude seems to be onto something.

* *"Schema pinned to 1.12.0 (majority convention in this repo; 1.28.0 also appears but is less common) instead of 1.28.0."*

I used 1.28.0 for quite some time since I liked that it was a proof-of-concept that it was even possible to use 1.28.0; however, I have (temporarily?) fell back to using 1.12.0 to save a couple minutes per package when re-submitting pkgs PRs' manifests to extras.

* *"dropped the additional nb (Norwegian) locale for simplicity — happy to add it back if wanted."*

I'd be merrily in favour of adding it back. There's not *all* that many apps out there that have distinctly separate official Norwegian names (Practically the entire population of Norway can read English, so almost all non-Microsoft or -Apple apps use the English names), so I grab the few opportunities I can get when I see apps that do have ones.

* *"Fixed a typo in the proposed tags (`system-center-storage-manager-backup` (…)"*

Good point.

shards/json/Microsoft.SystemCenter.ServiceManager.Backup.json
pl4nty, line 1: use the strategy from https://github.com/UnownPlain/anthelion/blob/main/CONTRIBUTING.md

pl4nty: I believe this already is the CONTRIBUTING.md-documented approach for this case — checking again to make sure I'm not missing something better:

- The URL (`download.microsoft.com/download/6314.../SCSM_2025.zip`) has no version anywhere in the path, and I couldn't find a Download Center listing page for this specific GUID (it doesn't appear indexed under `microsoft.com/en-us/download`), so `redirect-match`/`page-match`/`json` have no page or redirect target to read a version from.
- CONTRIBUTING.md's strategy table lists `static` for exactly this case ("No upstream version lookup is needed; pass a fixed value or installer metadata selector directly to komac"), with `nestedInstallerMatches` + `version: {source: "product"}` as the documented way to read the version from the installer's own file metadata inside an archive.
- This mirrors the sibling `Microsoft.SystemCenter.Orchestrator.RunbookDesigner` shard already merged in this repo, which hits the identical `download.microsoft.com` GUID-folder pattern (`SCO_2025.zip` instead of `SCSM_2025.zip`) and uses the same `static`/`product`/`nestedInstallerMatches` shape.

If you have a specific different strategy from the doc in mind for this one, let me know which and I'll switch — I don't want to guess further without more direction.

---

pl4nty: The GUID-folder pattern is not allowed, because it won't change. We need a different way to find the latest version, or a config.json change if it's discontinued

pl4nty: Good catch — I checked and there's genuinely no page or feed that surfaces a new GUID when Microsoft ships an updated build, so this can't be shard-covered honestly. Dropped the shard and added it to `repository/shard-coverage` in config.json instead (95ea077).

---

PR #1090, opened 2026-08-28
New package: ANBSoftware.TimeConverter version 1.1

pl4nty:
Checklist for Pull Requests

  - Relates to https://github.com/microsoft/winget-pkgs/pull/421322.

Manifests

- [x] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)? (manifests use the 1.12.0 schema, matching most of this repo's current manifests)

---

This relates to/replicates #1062 ("New package: ANBSoftware.TimeConverter version 1.1"), originally opened by @DandelionSprout, who noted they couldn't get an automated-update shard working because "The InstallerUrl is non-GitHub." This PR adds the same package plus a working shard, as a separate submission so #1062 is left untouched.

**Verification performed (no Windows environment available here, so `winget validate`/`winget install` could not be run):**

- Re-downloaded the installer archive directly from `anbsoftware.co.uk` and computed SHA256 myself (did not trust the hash quoted in #1062) — matched exactly: `3A6C43E07E73460396FBEA896CAEAC4337A622C71D22AAF0C3B503271FC4E398`.
- Confirmed 1.1.0.0 is still the current upstream version (product page still advertises "Latest version: v1.1.0.0").
- Extracted the MSI and verified its `ProductCode` (`{881F786E-9FF6-4AB4-85B7-227BEED7F747}`) and `UpgradeCode` with `msitools`' `msiinfo export` — confirms the ProductCode used in the manifest is correct.
- Ran this repo's own tooling against the new files (with `anthelion`'s transitive deps pinned directly in a scratch `package.json`, since the real `github:` dependency needs GitHub API access unavailable in this sandbox — `package.json`/`bun.lock` were restored afterward and are not part of this diff):
  - `bun manifests:check --deny-warnings` → `Manifests OK`
  - `bun test:manifests --deny-warnings` → all pass
  - `bun fmt --check` → clean for all files in this PR
  - Called `@unownplain/anthelion-komac`'s `analyzeInstaller` directly (the same native binding the CI `Test` job uses) against the shard's resolved URL to confirm it downloads, extracts, and reports the expected architecture/ProductCode/hash before pushing.

**Shard**: `shards/json/ANBSoftware.TimeConverter.json` uses `page-match` against the product page, which prints `Latest version: v1.1.0.0`. The regex captures the `Major.Minor` pair and reconstructs the known `{version}.0.0` filename convention both the version string and download filenames currently follow.

Two things I ran into and fixed along the way, worth flagging for review:

1. **komac's installer analyzer requires a real file extension in the URL path.** `anbsoftware.co.uk` serves downloads from `download.php?file=name.zip` (query string only), which the analyzer rejects (`Invalid file extension: php`) before even downloading. The server also honors PHP `PATH_INFO`, so `download.php/name.zip?file=name.zip` serves the byte-identical file (verified matching sha256/Content-Type/Content-Disposition) while giving the analyzer a real `.zip` to key off. The shard uses this form; I left the manifest's own `InstallerUrl` as the plain, simpler form DandelionSprout used in #1062 since it's fine for `winget install` itself (which reads `Content-Type`, not the URL), and future automated updates from the shard will use the working form.
2. **Dropped the exe-bootstrapper and portable-zip installer entries** that #1062 included alongside the MSI. All three shared the same `(Architecture=x86, Scope=none, effective InstallerType=zip)` tuple, since `NestedInstallerType` isn't part of this repo's installer-matching key anywhere: Validate's per-installer `concurrency.group` is keyed on that tuple, so `cancel-in-progress` cancelled two of the three matrix jobs every run (reproduced in CI on an earlier commit of this PR); and `validate.ps1` selects the installer to test with `Where-Object {...} | Select-Object -First 1`, so even without the cancellation it would only ever have exercised the MSI entry. Rather than patch the shared validate workflow/script (used by every other manifest in the repo) to disambiguate on `NestedInstallerType`, I kept just the MSI, which is the standard, most robust install method (proper Add/Remove Programs entry, ProductCode-based upgrades/uninstalls) — a real capability regression from #1062 for anyone who wanted the portable build, but it's what fits this repo's current tooling.


pl4nty:
`All Checks Pass` fails, but not because of a manifest problem — every individual check actually succeeds. Two of the three `Validate / Installation Validation (x86, ...)` matrix jobs come back `cancelled`, and the aggregate gate treats any cancellation as a failure.

Root cause: `.github/workflows/validate.yml`'s per-installer concurrency group is keyed on `${{ ref }}-${{ matrix.installer.path }}-${{ matrix.installer.arch }}-${{ matrix.installer.scope }}-${{ matrix.installer.type }}` (line 100). This manifest has three installer entries at the same path, same `Architecture: x86`, same (unset) `Scope`, and the same top-level `InstallerType: zip` (they only differ by `NestedInstallerType`: msi / exe / portable) — so all three matrix jobs resolve to the *identical* concurrency group, and `cancel-in-progress: true` kills the earlier two as soon as the third starts, entirely within the same run.

I confirmed this is deterministic, not a flake: I re-ran the failed jobs once (`rerun_failed_jobs`) and got the exact same 1-success/2-cancelled split again. A further re-run wouldn't help — this reproduces for any manifest with multiple installer entries sharing type/arch/scope, which is a legitimate, common manifest shape (this repo's own `Microsoft.SystemCenter.*` and portable+installer combos would be equally exposed). I don't think this is something I should fix as part of a new-package PR (it's shared CI workflow config, not this manifest), so flagging it for you to decide whether the concurrency key should also include something like the matrix index or `NestedInstallerType`/`InstallerLocale`.

---

pl4nty:
Update: superseded by later work here. There's now a real working shard (`page-match` against the product page, using a `download.php/<name>.zip?file=<name>.zip` PATH_INFO form to give the installer analyzer a recognized extension) rather than the "not feasible" ignore entry from my earlier comment. The concurrency-group collision I flagged is sidestepped by dropping the exe-bootstrapper and portable-zip installer entries down to just the MSI, so no two installers share the same `(arch, scope, type)` tuple — a real feature reduction versus #1062, but avoids depending on a shared-workflow fix. CI is fully green now, including the shard-analysis test.

---

DandelionSprout:
Claude's summaries are getting *really* wordy by now, so there's likely a lot of things I'll fail to spot that I'd normally be qualified to add context for.

* *"Dropped the exe-bootstrapper and portable-zip installer entries"*

That the .exe was a bootstrapper all along, does make sense now that I think about it. I am however not immediately sold on removing the `portable` installer, though that's because I like the concept of fallbacks, and that I know that there's a relatively small yet hardcore subcategory of PC nerds who want as many of their apps as possible to be portable (I am however not really in that subcategory).

* *"komac's installer analyzer requires a real file extension in the URL path."*

I suppose that counts as a feature/bug report that could be made to one or both of the main Komac distributions… I think.

* *"I confirmed this is deterministic, not a flake: (…)"*

I'm tempted to assume that "flake" was a typo and that Claude meant "fluke", but Aİs don't seem susceptible to make typos of pretty much anything, so I'm not sure what to think.

PR #868, opened 2026-08-14
New package: Mozilla.Firefox 153.0.3

DandelionSprout:
Checklist for Pull Requests

- [ ] Is there a related issue or pull request in [winget-pkgs](https://github.com/microsoft/winget-pkgs)? If so, add the link(s) below.
   
  - Relates to microsoft/winget-pkgs#[Issue Number]

Manifests

- [x] Have you checked that there aren't other open [pull requests](https://github.com/pl4nty/winget-extras/pulls) for the same manifest update/change?
- [x] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [x] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)?

Note: `<path>` is the directory's name containing the manifest you're submitting.

---
Notes from me:
* Resolves https://github.com/pl4nty/winget-extras/pull/577.
* Resolves https://github.com/pl4nty/winget-extras/pull/760.
* I tested out the official MSI installers too, but Winget softlocked when I tested them, as Winget failed to register that their installation was complete.
* I am not making a/the shard for this. Trust me on that.
* I should probably submit a bug report to winget-cli regarding a glitch I discovered that broke the prospect of using rare locales as InstallerLocale.

PR #760, opened 2026-08-04
New package: Mozilla.Firefox.Portable.RC version 153.0.3-candidates

DandelionSprout:
Checklist for Pull Requests

- [ ] Is there a related issue or pull request in [winget-pkgs](https://github.com/microsoft/winget-pkgs)? If so, add the link(s) below.
   
  - Relates to microsoft/winget-pkgs#[Issue Number]

Manifests

- [x] Have you checked that there aren't other open [pull requests](https://github.com/pl4nty/winget-extras/pulls) for the same manifest update/change?
- [x] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [x] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)?

Note: `<path>` is the directory's name containing the manifest you're submitting.

---
Notes from me:
* Based on #577, but does not resolve it.

PR #1093, opened 2026-08-28
New package: Mozilla.Firefox 154.0.1

pl4nty:
Checklist for Pull Requests

  - Relates to microsoft/winget-pkgs (Firefox is also published there; this is a winget-extras-specific build like #868 below)

Manifests

- [x] Have you checked that there aren't other open [pull requests](https://github.com/pl4nty/winget-extras/pulls) for the same manifest update/change? (#868 adds the same package at 153.0.3, without a shard — see notes below)
- [ ] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`? I don't have a Windows environment in this sandbox, so I could not run actual `winget validate`/`winget install`. Instead I verified with this repo's own tooling:
  - `bun manifests:check --deny-warnings` → `Manifests OK (2211 files).`
  - `bun test:manifests` (`bun test scripts`) → `89 pass, 0 fail`
  - `bun fmt --check` → clean
  - This repo's own CI (`winget install` on real Windows runners) did end up exercising every installer in this manifest — see the CI note below.
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`? Same caveat as above — no Windows runner available here, not run locally, but validated by this repo's CI.
- [x] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)? Uses schema 1.12.0, matching this repo's current convention (see other manifests under `manifests/`).

Note: `<path>` is `manifests/m/Mozilla/Firefox/154.0.1`.

---

## What this is

This replicates #868 ("New package: Mozilla.Firefox 153.0.3", opened by @DandelionSprout, who noted in that PR's own description that it "Resolves #577" and "Resolves #760") **as a separate PR**, updated to the current upstream version and — the part @DandelionSprout explicitly opted out of ("I am not making a/the shard for this. Trust me on that.") — with a working automated-update shard added. This PR does not touch, comment on, or close #868, #577, or #760; those are left for their author/maintainers to handle independently.

## Manifest

- `manifests/m/Mozilla/Firefox/154.0.1/` — MSIX, Nullsoft `.exe`, and a portable `zip`+`NestedInstallerType: portable` variant built from the MSIX payload, for x86/x64/arm64 (9 installers total), against **154.0.1**, which was `LATEST_FIREFOX_VERSION` per Mozilla's official [`firefox_versions.json`](https://product-details.mozilla.org/1.0/firefox_versions.json) at the time of writing (153.0.3 already exists upstream in microsoft/winget-pkgs, which is exactly the `repository/upstream-versions` warning DandelionSprout hit on #868).
- All hashes were independently verified rather than trusted from any prior PR:
  - The `en-US`/`win64` Nullsoft installer was downloaded directly with `curl` and hashed with `sha256sum`; it matched Mozilla's own published [`SHA256SUMS`](https://ftp.mozilla.org/pub/firefox/releases/154.0.1/SHA256SUMS) for the release exactly, so the `en-US` `x86`/`arm64` hashes were taken from that same official file.
  - The three MSIX packages (`win32`/`win64`/`win64-aarch64`, not listed in `SHA256SUMS`) were downloaded and hashed directly. `SignatureSha256` was computed by extracting `AppxSignature.p7x` from each MSIX (they're zip containers) and hashing that file directly. `PackageFamilyName` was left as `Mozilla.MozillaFirefox_jag0gd4e3s9p2`, confirmed unchanged by extracting `AppxManifest.xml` and checking the `Identity Name`/`Publisher` (the two inputs that determine it) match #868's package exactly.
- **Scope note vs. #868**: #868's manifest carried ~81 locales × 3 architectures of Nullsoft installers (plus MSIX/portable). This PR intentionally ships only `en-US` for the Nullsoft installer. I first built the full multi-locale manifest (with all locale hashes taken directly from Mozilla's official `SHA256SUMS`, using the same rare-locale exclusions DandelionSprout found in the #868 comment thread), pushed it, and watched this repo's own CI actually run it — see below for why that doesn't work here.

## CI finding: the concurrency-group / locale interaction

`.github/workflows/validate.yml` keys each `Installation Validation` matrix job's `concurrency.group` by `matrix.installer.path-arch-scope-type`, which does **not** include `InstallerLocale`. With dozens of locale variants sharing the same architecture/scope/type, `cancel-in-progress: true` collapses them onto each other within a single run — only the last-started job per `(arch, scope, type)` survives, and the rest complete as `cancelled`. The repo's `All Checks Pass` gate then hard-fails on any cancelled job (`CANCELLED: true` → `exit 1`), regardless of `fail-fast: false` on the matrix itself.

I confirmed this on the first push of this PR: 243 of 246 Nullsoft jobs came back `cancelled` (not `failure`) and only one per architecture actually ran `winget install`, which is exactly what triggered `All Checks Pass` to fail. This isn't specific to my manifest — it would affect #868's manifest identically, since it has the same many-locales-per-architecture shape.

To get a genuinely green CI run rather than a manifest that only looks complete, I reduced this PR's manifest to `en-US` only for the Nullsoft installer (3 architectures × MSIX/Nullsoft/portable = 9 installers, no two sharing an `(arch, scope, type)` tuple). All 9 real installer-validation jobs now pass.

## Shard

`shards/json/Mozilla.Firefox.json` — a `json`-strategy shard (per [anthelion's CONTRIBUTING.md](https://raw.githubusercontent.com/UnownPlain/anthelion/main/CONTRIBUTING.md)), which turned out to be very achievable despite the "trust me" in #868:

- Version source: Mozilla's official `https://product-details.mozilla.org/1.0/firefox_versions.json`, path `LATEST_FIREFOX_VERSION` — a clean, single version string, no parsing needed.
- `urls`: one templated URL per installer (3 MSIX + 3 Nullsoft `en-US`; the portable variant reuses the MSIX URLs so isn't duplicated), built from Mozilla's stable `https://download-installer.cdn.mozilla.net/pub/firefox/releases/{version}/<arch-dir>/<locale>/Firefox%20Setup%20{version}.exe` (and `.../multi/Firefox%20Setup%20{version}.msix`) pattern, confirmed against real directory listings.
- Validated the shard file itself against anthelion's actual JSON schema (`https://anthelion.unownplain.dev/schema.json`) with `ajv` — passes (`valid: true`). I could not run `bun test:shard`/`anthelion-test` end-to-end because the real `anthelion` package needs GitHub API access this sandbox's token doesn't have; `bun manifests:check` (which only checks that a shard file exists, not its internal correctness) passes and no longer reports the `repository/shard-coverage` warning.

## AI disclosure

This PR (manifest generation, hash verification, and shard) was produced with heavy use of Claude Code, per this repo's normal contribution norms — flagging this explicitly since the linked anthelion project's own CONTRIBUTING.md asks contributors to disclose AI involvement.


pl4nty:
`All Checks Pass` fails here for the same reason I diagnosed on #1090: `.github/workflows/validate.yml`'s per-installer concurrency group (line 100) doesn't include `InstallerLocale`, so every locale variant that shares path/arch/scope/type collapses into one concurrency group and `cancel-in-progress: true` cancels all but one — that's the ~240+ `cancelled` entries here, not a manifest defect (the real, surviving jobs for each install-type all show `success`). Since this is deterministic (verified by reproducing it identically on a re-run of #1090), I'm not re-running this one too — same root cause, same expected outcome. See my comment on #1090 for the full diagnosis; not duplicating it here.

---

manifests/m/Mozilla/Firefox/154.0.1/Mozilla.Firefox.installer.yaml
```diff
@@ -0,0 +1,146 @@
...
+  NestedInstallerFiles:
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\firefox.exe
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\crashhelper.exe
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\crashreporter.exe
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\default-browser-agent.exe
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\nmhproxy.exe
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\pingsender.exe
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\plugin-container.exe
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\private_browsing.exe
+  - RelativeFilePath: VFS\ProgramFiles\MozillaFirefox%20Package%20Root\desktop-launcher\desktop-launcher.exe
```
pl4nty: I don't think these belong on PATH

manifests/m/Mozilla/Firefox/154.0.1/Mozilla.Firefox.locale.en.yaml
```diff
@@ -0,0 +1,42 @@
...
+LicenseUrl: https://www.mozilla.org/MPL/2.0
+Copyright: © Firefox and Mozilla Developers.
+CopyrightUrl: https://www.mozilla.org/foundation/trademarks/policy
+ShortDescription: The legendary web browser by Mozilla.
+Description: |-
+  The legendary web browser by Mozilla.
+
+  This winget-extras package bundles three installer types for x86, x64, and arm64: the official MSIX package, the Nullsoft (.exe) installer, and a portable build derived from the MSIX payload.
+
+  With stock Winget settings, fresh installs use the Nullsoft .exe installer. To install the MSIX version instead, use "winget install Mozilla.Firefox --installer-type msix -s winget-extras".
```
pl4nty: I don't think this is correct anymore

manifests/m/Mozilla/Firefox/154.0.1/Mozilla.Firefox.installer.yaml
pl4nty: Right — for `NestedInstallerType: portable`, every `NestedInstallerFiles` entry gets its own PATH symlink, so listing all 9 companion binaries was exposing internal helpers (`crashhelper.exe`, `pingsender.exe`, etc.) as top-level commands. Pushed a fix keeping only `firefox.exe` (with `PortableCommandAlias: firefox`) across all three architectures; `ArchiveBinariesDependOnPath: true` still keeps the rest of the extracted directory in place for `firefox.exe` to find its siblings at runtime.

---

manifests/m/Mozilla/Firefox/154.0.1/Mozilla.Firefox.locale.en.yaml
pl4nty: Fair — I wasn't confident in that claim about winget's default installer-type selection to begin with. Pushed a fix that drops the specific claim and just tells users how to pick a type explicitly.

---

PR #1159, opened 2026-09-03
 refactor(linter): migrate manifest schemas to Zod

UnownPlain:
Migrates the manifest linter to use Zod to remove some dependencies. Zod 4.5 now supports [AOT compilation](https://zod.dev/compile) so performance should be comparable. The downside is that schemas can't be auto generated from upstream JSON schemas but that shouldn't be an issue since the schema doesn't change much between versions.


pl4nty (approved):
I'd usually prefer codegen'd types, but it's your code and the schemas don't change much. ty!

PR #1169, opened 2026-09-04
New package: Microsoft.Sysinternals.Suite version 2026-08-19

UnownPlain:
This package never passes validation upstream for some reason which causes updates to be delayed for multiple weeks given the number PRs open upstream and there being only 1 active admin.

https://github.com/microsoft/winget-pkgs/pull/421013


pl4nty:
thanks! these delays are getting to be a headache...

PR #1244, opened 2026-09-08
New packages: five winget-pkgs requests that can't be added upstream

pl4nty:
Adds five packages triaged from open `microsoft/winget-pkgs` issues and pull requests that are blocked upstream.

Checklist for Pull Requests

- [x] Is there a related issue or pull request in [winget-pkgs](https://github.com/microsoft/winget-pkgs)? If so, add the link(s) below.
  - Relates to microsoft/winget-pkgs#421811 (`0xJacky.nginx-ui`, Binary-Validation-Error)
  - Relates to microsoft/winget-pkgs#429325 (`go-gost.Gost`, Binary-Validation-Error)
  - Relates to microsoft/winget-pkgs#419190 (`k0sproject.k0sctl`, Binary-Validation-Error)
  - Relates to microsoft/winget-pkgs#414760 (`Uotan-Dev.UotanToolboxNT`, Binary-Validation-Error)
  - Relates to microsoft/winget-pkgs#35157 (`ESET.OnlineScanner`, Interactive-Only-Installer)

Manifests

- [x] Have you checked that there aren't other open [pull requests](https://github.com/pl4nty/winget-extras/pulls) for the same manifest update/change?
- [ ] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`?
- [x] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)?

---

## Packages

| Package | Version | Upstream blocker |
| --- | --- | --- |
| `0xJacky.nginx-ui` | 2.5.10 | Binary validation false positive |
| `go-gost.Gost` | 3.3.0 | Binary validation false positive |
| `k0sproject.k0sctl` | 0.32.2 | Binary validation false positive |
| `Uotan-Dev.UotanToolboxNT` | 3.7.1 | Binary validation false positive |
| `ESET.OnlineScanner` | 3.7.11.0 | Interactive-only installer |

## Notes for reviewers

- All manifests were generated with the Anthelion fork of komac 2.16.0 from the real installer bytes; every SHA256 comes from a download, none is hand-written.
[...]

PR #1253, opened 2026-09-08
New package: Nvidia.Broadcast version 2.2.1

pl4nty:
Checklist for Pull Requests

  - No upstream request exists yet. Raised on pl4nty/winget-extras#1244 as a hardware-gated candidate; winget-pkgs carries its predecessor `Nvidia.RTXVoice` but not Broadcast.

Manifests

---

Requires an RTX GPU to install, so it cannot be verified upstream.

## Installation Validation cannot pass here, by construction

The runner has no RTX GPU, so the installer exits `3858759680` and there is no manifest change that alters that. Two runs establish it:

1. Before the fix, winget exited `0x8a150006` (generic install failure).
2. With `ExpectedReturnCodes` mapping `3858759680` → `systemNotSupported`, winget now exits `0x8a150113` (`SYSTEM_NOT_SUPPORTED`) — the mapping works and winget classifies the outcome correctly.

`validate.ps1` fails on any non-zero winget exit, so the check stays red either way. The mapping is kept because it is correct for real users on real hardware: they get "system not supported" instead of an opaque installer failure. That code came from CI, not from guesswork.

This is not specific to this PR. Every hardware-gated NVIDIA package in this repo hits it, including the ones this repo's own updater opens: pl4nty/winget-extras#1380, pl4nty/winget-extras#1381 and pl4nty/winget-extras#1382 each carry the same `ExpectedReturnCodes` → `systemNotSupported` pattern as the `Nvidia.Driver.*` manifests already on `main`, and each has Installation Validation red for the same reason. The common cause is in `validate.ps1`:

```powershell
if ($installer.ExitCode -ne 0) {
    throw "Install failed with exit code $($installer.ExitCode)"
}
```

It compares winget's raw exit code against zero, so a manifest that correctly declares an expected return code still fails. Honouring the manifest's `ExpectedReturnCodes` there would turn this class green without weakening validation for anything else. That is a shared-CI change affecting every package, so it is left for a maintainer rather than made from this PR.

## Other notes

The installer fields follow this repo's existing `Nvidia.NVIDIAApp` manifest: `InstallerType: exe`, `Silent: -s`, `elevationRequired`, `Scope: machine`, `Architecture: x64`. Komac classifies the download as `portable` because it does not recognise NVIDIA's self-extracting stub, and reads that stub's 32-bit PE header as `x86`; both are corrected by hand to match the sibling package.

`ProductCode` is still not set — it cannot be read without a successful install on RTX hardware.

The shard captures the whole download URL, as `Nvidia.Nsight.Systems` does, because the URL carries both the `2.2.1` directory and the `2.2.1.58338310` build. The regex was checked against the live page: one match, resolving to version `2.2.1`.


DandelionSprout:
The ProductCode from what I can tell is `{B2FE1952-0186-46C3-BAEC-A80AA35AC5B8}_NVIDIABroadcast`

This PR is also of interest to:
* https://github.com/microsoft/winget-pkgs/pull/298144
* https://github.com/microsoft/winget-pkgs/issues/220093
* https://github.com/microsoft/winget-pkgs/pull/221618

PR #1256, opened 2026-09-08
New package: rgst-io.stencil version 3.0.0

pl4nty:
Adds stencil, a Go project scaffolding and templating tool.

Relates to microsoft/winget-pkgs#422240.

- Manifest generated with komac from the v3.0.0 release assets, carrying all three Windows architectures (x86, x64, arm64) in one manifest.
- `Commands: [stencil]` so the tool is on PATH after install.

**Shard corrected in f92ad8f following review.** I had used `page-match` with a regex over the raw body of a Forgejo releases API response. Anthelion has a dedicated `json` strategy for "APIs or metadata files containing version", which reads it with a dot-separated path instead:

```json
{
  "strategy": "json",
  "json": {
    "url": ".../api/v1/repos/rgst-io/stencil/releases?limit=1&draft=false&pre-release=false",
    "path": "0.tag_name"
  }
}
```

This follows the existing `GarudaLinux.FireDragon` shard, which is the same shape — `0.tag_name` against a forge releases API where tags carry a `v` prefix, with `v{version}` in the tag path and bare `{version}` in the filename. The URLs are unchanged.

Worth noting this is `git.rgst.io`, a self-hosted Forgejo instance rather than github.com, so `github-release` does not apply here.

**I also re-swept every shard in my open PRs for the same mistake.** My first sweep after the Bluebook review only checked URLs whose path ended in `.yml`/`.yaml`/`.json`, which is precisely why it missed this one — the URL ends in query parameters. I redid it by fetching each `page-match` target and checking its `Content-Type`: of 28 shards, 13 use `page-match`, and all 13 remaining ones return `text/html`. Bluebook and stencil were the only two aimed at machine-readable feeds, and both are now on dedicated strategies.

This push also merges `main`.


shards/json/rgst-io.stencil.json
```diff
@@ -0,0 +1,13 @@
+{
+	"$schema": "https://anthelion.unownplain.dev/schema.json",
+	"strategy": "page-match",
+	"pageMatch": {
+		"url": "https://git.rgst.io/api/v1/repos/rgst-io/stencil/releases?limit=1&draft=false&pre-release=false",
+		"regex": "\"tag_name\"\\s*:\\s*\"v(\\d+(?:\\.\\d+)+)\""
+	},
```
UnownPlain: Read the docs

UnownPlain (approved):

PR #1257, opened 2026-09-08
New package: Cloudflare.Warp.Beta version 2026.8.1290.1

pl4nty:
Adds the beta channel of the Cloudflare One Client (WARP).

Relates to microsoft/winget-pkgs#190056.

**Reworked in 7ab8a28 to use the beta endpoint, as suggested in review.**

`https://downloads.cloudflareclient.com/v1/download/windows/beta` is a stable URL that 302s to the versioned installer:

```
location: https://downloads.cloudflareclient.com/v1/download/windows/version/2026.8.1290.1
```

That is exactly what Anthelion's `redirect-match` strategy is for — "stable URLs redirecting to versioned installers" — so the shard now uses it:

```json
{
  "strategy": "redirect-match",
  "redirectMatch": {
    "url": ["https://downloads.cloudflareclient.com/v1/download/windows/beta"],
    "regex": "/windows/version/(\\d+(?:\\.\\d+)+)"
  }
}
```

`urls` is omitted deliberately: per the docs, the redirect destination is then handed to komac, which is the versioned installer URL we want. Verified against the live endpoint — the regex captures `2026.8.1290.1` from the `Location` header.

This replaces a `page-match` regex over the beta-releases documentation page, which was the wrong tool and would break on any docs restyle.

**The version is also bumped, which the switch exposed.** The endpoint now serves `2026.8.1290.1`; the PR previously pinned `2026.7.1210.1`. Manifest regenerated with komac, so the hash, `ProductCode` and `DisplayVersion` are all refreshed. Note the ProductCode changes between builds (`{C3AD24EA-…}` → `{399DC79E-…}`) while the UpgradeCode `{1BF42825-7B65-4CA9-AFFF-B7B5E1CE27B4}` stays constant, which is the usual WiX arrangement.

One thing to be aware of with this strategy: because the beta channel always resolves to the newest build, this package will track it continuously rather than sitting on a pinned version.

The `Hashicorp.Terraform` lint failure noted here earlier is gone — #1410 landed the ignore, and this branch has since merged `main`.


pl4nty (commented):
use https://downloads.cloudflareclient.com/v1/download/windows/beta to get the latest version

PR #1263, opened 2026-09-08
New package: LDPlayer.LDPlayer.14 version 14.0.25.1 and LDPlayer.LDPlayer.9 version 9.5.32.0

pl4nty:
Checklist for Pull Requests

  - Relates to microsoft/winget-pkgs#153615

---

Blocked upstream as an interactive-only installer.

Split per review into the two majors the vendor maintains concurrently:

| Package | Version | Released | Size |
| --- | --- | --- | --- |
| `LDPlayer.LDPlayer.14` | 14.0.25.1 | 2026-08-28 | 972,683,232 |
| `LDPlayer.LDPlayer.9` | 9.5.32.0 | 2026-08-05 | 795,967,968 |

The unsuffixed `LDPlayer.LDPlayer` package and its shard are removed in the same commit — it was never published, so nothing is orphaned.

### Where the versions come from

The release notes page carries a separate record list per major stream (`versionFourteen`, `versionNine`, and older `versionFive`/`versionFours`/`versionThrees`), each record holding `version`, `releaseDate` and a version-pinned `link`. Each shard anchors its regex to its own stream key and takes the first record, so the 14 shard can never pick up a 9 release or vice versa. The download URL is taken from the matched record rather than rebuilt from the version, because the host varies between records (`setup.ldrescdn.com` on the current ones, `static.ldrescdn.com` on older ones).

Both shards were run against the live page and resolve to exactly the versions and URLs in these manifests.

### Corrections to the previous revision of this PR

- It claimed LDPlayer 9 publishes no version number. That was wrong — the release notes page has always carried the `versionNine` records, and they are the source used here.
- It described `LDPlayer_14.0.25.2.exe` as coming from the release notes page. It does not appear there; the newest 14 record is 14.0.25.1. 14.0.25.2 does exist on the CDN (972,687,336 bytes, sha256 `1C2E3448…`) but is unannounced, so packaging it would put a version in the repo that the shard could never reproduce. Say the word if you would rather carry the unlisted build.

### Manual corrections to komac's output

Both installers are custom stubs with no Inno/NSIS/Burn marker, so komac calls them `portable` and reads `x86` from the stub's PE header (both stubs are genuinely i386 — verified from the `IMAGE_FILE_HEADER` machine field). Corrected by hand to `exe` and `x64`, LDPlayer being a 64-bit emulator. These are the full offline packages, not the ~3 MB web installer the download button serves.

`InstallModes: [interactive]` is set because the installer is a click-through wizard, which is what the upstream label refers to — not to dodge a CI timeout. Both Installation Validation jobs time out on the installer, which is the expected result for that label.

### CI

Green. The first `LDPlayer.LDPlayer.9` attempt failed before it reached the installer — `validate.ps1` could not install the `powershell-yaml` module, five attempts over ~40 s, each `No match was found for the specified search criteria and module name 'powershell-yaml'` alongside a NuGet CLI `Missing option value for: '-source'`. The sibling `LDPlayer.LDPlayer.14` job on the same run cleared that step and passed, so it was a transient PSGallery/NuGet fault on that one runner. One re-run confirmed it: attempt 2 passed. Recorded here only so a future reader of the run history knows what that red job was.

Both packages are in one PR because they share a single source page and the split is one change; happy to break the 9 out into its own PR if you prefer one package per PR.

### DIFF
--- a/manifests/l/LDPlayer/LDPlayer/14.0.25.2/LDPlayer.LDPlayer.installer.yaml
+++ b/manifests/l/LDPlayer/LDPlayer/14.0.25.2/LDPlayer.LDPlayer.installer.yaml
@@ -1,0 +1,14 @@
+# Created by Anthelion using komac v2.16.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: LDPlayer.LDPlayer
+PackageVersion: 14.0.25.2
+InstallerType: exe
+InstallModes:
+- interactive
+Installers:
+- Architecture: x64
+  InstallerUrl: https://setup.ldrescdn.com/download/package/LDPlayer_14.0.25.2.exe
+  InstallerSha256: 1C2E3448B3837CBC770AABC7ACED20EA48C4E518B78DFFF8B6854AC639B93EF9
+ManifestType: installer
+ManifestVersion: 1.12.0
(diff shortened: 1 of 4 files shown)


pl4nty (commented):
Major versions are maintained concurrently, so we should add `LDPlayer.LDPlayer.14` and `LDPlayer.LDPlayer.9`. Version can come from the release notes page

PR #1346, opened 2026-09-08
New package: xupefei.Locale-Emulator version 2.5.0.1

pl4nty:
Adds Locale Emulator, which runs applications under a different system locale.

Fixes microsoft/winget-pkgs#332489 — declined upstream under the `Interactive-Only-Installer` label.

- Manifest generated with komac from the v2.5.0.1 release asset.
- Licence read from the source repo, which ships both `COPYING` (GPLv3) and `COPYING.LESSER` (LGPLv3) — the standard FSF layout for an LGPL project, so the manifest uses `LGPL-3.0-only`.
- Shard uses the `github-release` strategy against `xupefei/Locale-Emulator`.

Checked for an existing package before building: no match by identifier, by name token, or by source repository (`github.com/xupefei/Locale-Emulator`) in either winget-pkgs or winget-extras.

The archive holds four executables, so all four are listed as `NestedInstallerFiles`; `LEGUI.exe` is the entry point a user would launch.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cg5ocVwbGciXXmwpZTu9Ry)_

### DIFF
--- a/manifests/x/xupefei/Locale-Emulator/2.5.0.1/xupefei.Locale-Emulator.installer.yaml
+++ b/manifests/x/xupefei/Locale-Emulator/2.5.0.1/xupefei.Locale-Emulator.installer.yaml
@@ -1,0 +1,19 @@
+# Created by Anthelion using komac v2.16.0
+# yaml-language-server: $schema=https://aka.ms/winget-manifest.installer.1.12.0.schema.json
+
+PackageIdentifier: xupefei.Locale-Emulator
+PackageVersion: 2.5.0.1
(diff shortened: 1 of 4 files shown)


DandelionSprout:
What's interesting about this specific app, is that it was hard for me during preliminary testing earlier this year to determine what counted as installed or not.

The thing was (abbreviated) that after running the app, a notice is shown that claimed the notice could immediately be closed and that *that* was the point where the app itself considered itself installed (as user-scope).

PR #1353, opened 2026-09-08
New package: 3u.3uTools version 9.08.006

pl4nty:
Adds 3uTools, which manages iOS device files, firmware and backups.

Fixes microsoft/winget-pkgs#118743 — declined upstream under the `Interactive-Only-Installer` label.

- Manifest generated with komac from the vendor downloads, carrying both architectures (x86 and x64) in one manifest.
- Licence is `Proprietary`: no SPDX licence, distributed free under the vendor's own terms.

Checked for an existing package before building: no match by identifier, by word-subset name match, or by source host (`dl.3u.com`) in either winget-pkgs or winget-extras. The only near-hits anywhere were unrelated m3u8 tools.

**Installer type — changed from komac's output.** komac detected `portable`; left that way winget would symlink a 200MB setup binary rather than run it. The files are published as `3uTools_v9.08.006_Setup_x64.exe`, so they are installers.

I could not establish a silent switch: the binary carries no NSIS, Inno or InstallShield signature (sections are `.text .rdata .data .tls .rsrc .reloc`, a plain compiled binary with a custom setup UI). I set `InstallerType: exe` with `InstallModes: [interactive]` rather than guess a switch — accurate to the upstream rejection reason, and the case `validate.ps1` expects to time out.

I deliberately left `Scope` and `ElevationRequirement` unset: I could not read the resource section within the range I fetched, so asserting either would have been a guess.

**Shard corrected following review.** The `v900` channel segment was hardcoded, which would break the shard the moment 3uTools moves to a new channel directory. The download page carries the full URL, so the segment is now captured from it alongside the version:

```
https://dl\.3u\.com/update/(?<channel>v\d+)/dl/3uTools_v(?<version>\d+\.\d+\.\d+)_Setup_x64\.exe
```

with the URLs templated as `.../update/{captures.channel}/dl/...`. Verified against the live page: it resolves to `channel=v900`, `version=9.08.006`, reproducing both manifest URLs exactly. Nothing in the path is fixed now except the host and the `update`/`dl` segments.

This push also merges `main`, which the branch was 31 commits behind.


shards/json/3u.3uTools.json
```diff
@@ -0,0 +1,12 @@
+{
+	"$schema": "https://anthelion.unownplain.dev/schema.json",
+	"strategy": "page-match",
+	"pageMatch": {
+		"url": "https://www.3u.com/",
+		"regex": "3uTools_v(\\d+\\.\\d+\\.\\d+)_Setup_x64\\.exe"
+	},
+	"urls": [
+		"https://dl.3u.com/update/v900/dl/3uTools_v{version}_Setup_x64.exe",
+		"https://dl.3u.com/update/v900/dl/3uTools_v{version}_Setup_x86.exe"
```
UnownPlain: Don't think the 900 should be hardcoded

UnownPlain (approved):

PR #1354, opened 2026-09-08
New package: DevXDojo.MrRSS version 1.3.32

pl4nty:
Adds MrRSS, a desktop RSS feed reader.

Fixes microsoft/winget-pkgs#339507 — declined upstream under the `Interactive-Only-Installer` label.

- Manifest generated with komac from the v1.3.32 release assets, carrying both Windows architectures (x64 and arm64) in one manifest.
- Licence read from `LICENSE` in the source repo: GPL-3.0-only.
- Shard uses the `github-release` strategy and lists both installer URLs.
- komac's detection needed no adjustment — NSIS, `ProductCode: MrRSS`, the `Ch3nyang` ARP publisher entry and the install location all came through.

Checked for an existing package before building: no match by identifier, by word-subset name match, or by source repository in either winget-pkgs or winget-extras.

The upstream issue was raised against 1.3.20; this packages the current 1.3.32.

**Two CI failures, both fixed.**

`Test / Test` (fixed in bd54e1b). I originally used `WCY-dt/MrRSS`, the owner named in the upstream issue. The project has since moved to `DevXDojo/MrRSS`, and Anthelion's shard test rejects a moved repository outright:

```
Error: GitHub repository has moved: https://github.com/DevXDojo/MrRSS/releases/latest
```

The manifests, shard and `PackageIdentifier` now use the canonical owner. Installer hashes are byte-identical before and after, confirming the old URLs redirected to the same assets rather than different builds. Flagging that the identifier is therefore `DevXDojo.MrRSS`, not the `WCY-dt.MrRSS` the upstream issue proposed.

`Installation Validation (arm64)` (fixed in 35ef352). The x64 installer completes a silent install and passed. The arm64 build hangs under the same invocation — five-minute timeout, no exit code, on a genuine `windows-11-arm64` runner (the workflow routes arm64 jobs there, so this is not an emulation artefact). Same NSIS packaging, different behaviour per architecture.

`validate.ps1` resolves install modes per installer:

```powershell
$installModes = @($selectedInstaller.InstallModes ?? $manifest.InstallModes)
```

so I scoped `InstallModes: [interactive]` to the arm64 entry alone and left x64 installing silently, rather than degrading both architectures to interactive.

One caveat for the record: that arm64 declaration rests on a single observation. It is consistent and on real hardware, but if anyone has an arm64 machine and finds it does install silently, the entry should come back out.


pl4nty (commented):
it doesn't make sense that only arm64 would be interactive, and the screenshot is broken. fix it first

PR #1425, opened 2026-09-11
fix: seed missing version-state files for 5 shards

pl4nty:
## What I checked

- Most recent CI run on `main` (run [34644847097](https://github.com/pl4nty/winget-extras/actions/runs/34644847097)): fully green, 0 manifest-linter warnings/errors, no shard-coverage gaps.
- Most recent [Update Packages](https://github.com/pl4nty/winget-extras/actions/workflows/update-packages.yml) run (run [34642295107](https://github.com/pl4nty/winget-extras/actions/runs/34642295107)): 382/391 shards succeeded, 9 failed.

## What I fixed

Five of the nine shard failures were the same root cause: anthelion's state-diffing helper does an unconditional read of `version-state/<PackageIdentifier>` and throws `ENOENT` if the file doesn't exist yet (rather than treating a missing file as "no prior state"), so every scheduled run for these packages has been erroring instead of ever getting the chance to write that file itself:

- `TeXLive.TeXLive` (state: `last-modified` header)
- `Microsoft.SystemCenter.VirtualMachineManager.Client` (state: `etag`)
- `ESET.OnlineScanner` (state: `etag`)
- `Microsoft.AccessRuntime` (state: `etag`)
- `DocRepairProject.Agdasima.Font` (state: latest commit touching `fonts/ttf` in the upstream repo)

I seeded each `version-state/<id>` file with the value anthelion would compute right now (fetched directly from the same URLs/headers each shard already points at, or the latest commit for the Agdasima font path), matching the exact raw format (quotes included for etags) used by every other file in `version-state/`. I cross-checked every other shard that declares a `state` field against `version-state/` and confirmed there are no other gaps.

Verified locally: `bun test scripts`, `bun --experimental-http2-fetch scripts/manifest-linter`, and `bun --bun oxfmt --check` all pass.

## What I deliberately left alone

- `Microsoft.VisualStudioCode` warnings/state, per standing instructions (upstream winget-pkgs versions, tracked deliberately).
- `FreeFileSync.FreeFileSync` and `CnCNet.RedAlert2YurisRevenge`: both fail installer analysis with `Inno Setup version 7.0.0.3 is newer than the maximum supported version 6.7`. This is a limitation in the upstream `anthelion` installer-analysis library itself, not something fixable from a shard config in this repo. Since the current installer already ships that Inno Setup version, this will likely keep failing until anthelion adds support — flagging for visibility rather than disabling the (otherwise working) shards.
- `nba-emu.NanoBoyAdvance` and `Shinmera.PromptFont.Font`: both failed with a `TimeoutError` hitting the `codeberg.org` releases API. This looks like transient upstream flakiness rather than a repo bug (the shard config is a straightforward JSON API strategy that mirrors many working codeberg-backed shards), so no action taken; the next scheduled run should retry naturally.
- No `repository/shard-coverage` warnings existed on `main` at the time of this run, so there was nothing to add a shard (or `.disabled` fallback) for.

https://claude.ai/code/session_01BFQXSjwKQZLR1f5EHD547n

---
_Generated by [Claude Code](https://claude.ai/code/session_01BFQXSjwKQZLR1f5EHD547n)_


UnownPlain:
> FreeFileSync.FreeFileSync and CnCNet.RedAlert2YurisRevenge: both fail installer analysis with Inno Setup version 7.0.0.3 is newer than the maximum supported version 6.7. This is a limitation in the upstream anthelion installer-analysis library itself, not something fixable from a shard config in this repo. Since the current installer already ships that Inno Setup version, this will likely keep failing until anthelion adds support — flagging for visibility rather than disabling the (otherwise working) shards.

Believe this should be fixed

UnownPlain (approved):

pl4nty:
Dug into where the Inno Setup cap actually lives before proposing anything, since I want to avoid a change here that can't actually fix it:

The "Inno Setup version 7.0.0.3 is newer than the maximum supported version 6.7" message comes from the Rust native binary shipped as `@unownplain/anthelion-komac` (I found the string embedded in `anthelion-komac.linux-x64-gnu.node`), not from any JS/TS in this repo or in `anthelion-external`. `anthelion-external` pins that package at `0.0.65`, which is also the current `latest` on npm — so there's no newer release to bump to yet, and nothing in `winget-extras`' shard config can route around Komac's own installer parser.

So this genuinely isn't fixable from this PR or this repo: it needs Inno Setup 7.x support added to Komac's installer analyzer first, then a new `@unownplain/anthelion-komac` release, then a bump of the pinned `anthelion` commit here (Renovate already does that last part automatically, per #1424 and similar). Happy to open a tracking issue here if useful, but let me know if you'd rather track it against Komac directly since that's where the actual fix needs to land.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BFQXSjwKQZLR1f5EHD547n)_

---

PR #1444, opened 2026-09-13
fix: match locale manifest Publishers and PackageNames to ARP

pl4nty:
Follows the reviewer edits on #1441, where the locale `Publisher` became `Splashtop Inc.` and the ARP `Publisher` line was then removed.

**Skill** — one sentence in `new-package-pr`: prefer matching the locale `Publisher`/`PackageName` to the ARP `Publisher`/`DisplayName` and dropping the redundant `AppsAndFeaturesEntries` field, unless the ARP value is a worse name for users.

**Manifests** — a sweep of all 131 version dirs that carry an ARP `Publisher` or `DisplayName`. 31 dirs change:

- **23 dirs** lose a field that repeated the locale value byte for byte. No display change.
- **8 packages** get the locale name pointed at the vendor's own string first, which makes the override redundant the same way: `Buypass AS`, `FlexRadio, Inc.`, `OneLogin, Inc`, `Purslane Tech Pte. Ltd.`, `AzIP LogViewer`, `Configuration Manager Support Center`, `Orca`, `Samsung NVM Express Driver`. These land on the current version dir only, because older dirs still match the ARP string that shipped with them — RustDesk 1.4.6 and 1.4.8 write `PURSLANE`, for example.
- The remaining ~100 dirs keep their override, because the ARP value is the worse name: bare domains (`cncnet.org`, `getcomposer.org`, `audiveris.org`), `Microsoft` for `Microsoft Corporation`, internal codenames (`bannerApp`, `MAppsCli`, `sysapi`), an OEM for `MSI.TrueColor`, and every version- or arch-bearing `DisplayName` (all `Nvidia.Driver.*`, all `Microsoft.DotNet.*`).

**One manifest correction.** `FlexRadioSystems.SmartSDR` declared `InstallModes: [interactive]`, which makes `validate.ps1` require the install to hang; its MSI installed silently and exited 0 in three separate runs. The declaration was wrong, so it is gone, and both SmartSDR validation jobs now pass.

Edits are line-exact: each line keeps its own terminator, so the mix of LF and CRLF across the manifests survives and no file is rewritten unless a line in it changed.

### Validation

Locally: `bun manifests:check --deny-warnings` (2637 files OK), `bun test:manifests` (138 pass), `bun fmt`. A re-sweep confirms no redundant ARP name is left.

In CI on `64af017`: 43 of 52 validation jobs pass. The 9 failures are pre-existing conditions this PR only surfaces, because the matrix is built from *changed* manifests and these packages have not been re-validated since they were added — see the [diagnosis comment](https://github.com/pl4nty/winget-extras/pull/1444#issuecomment-5651478322). `OneLogin.DesktopPro` was proven pre-existing by dispatching `validate.yml` against `main`, where it fails identically.

| Still failing | Why |
| --- | --- |
| VCRedist arm32 ×3 | No ARM32 Windows runner exists; the PR carries `Validation-arm32` for this |
| Nvidia.Driver.GeForceGameReady ×3 | Installer `0xE0000242` — no GPU on hosted runners |
| OneLogin.DesktopPro | MSI 1603, needs a tenant subdomain |
| Samsung.NVMeDriver | Hangs past the 5 min limit with no Samsung NVMe device present |

`AppsAndFeaturesEntries` is read by winget after install, to correlate an installed package back to a manifest. It is never passed to the installer, so removing a redundant `Publisher` or `DisplayName` cannot change an install outcome — and the 43 passing installs cover every other package in the sweep.

Two follow-ups worth their own issues, not fixed here:

- `Samsung.NVMeDriver` hangs but does not declare `interactive`, the inverse of the SmartSDR case. Adding it would make validation expect the timeout, but one hardware-less runner is thin evidence that the installer always hangs.
- `validate.yml` could skip installation validation for metadata-only diffs, which would stop this recurring on every future metadata PR.

Checklist for Pull Requests

  - N/A - no upstream issue.

Manifests

- [ ] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`? — no Windows host; `bun manifests:check` and the CI matrix above instead.
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`? — no Windows host; 43 CI installs instead.
- [x] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)? — schema validation is part of the linter run above.

---

https://claude.ai/code/session_01NvPp9nnYAfHPRjZDH6hvHJ


pl4nty:
`Installation Validation` failed on `1d44a4a` for the three VCRedist arm32 packages. The cause is the workflow, not this diff:

```
Installer [Arm,burn,Unknown,] not applicable: Machine is not compatible with Arm
Terminating context: 0x8a150010   (APPINSTALLER_CLI_ERROR_NO_APPLICABLE_INSTALLER)
Install failed with exit code -1978335216
```

`validate.yml` picks the runner with `matrix.installer.arch == 'arm64' && 'windows-11-arm' || 'windows-latest'`, so `Architecture: arm` lands on an x64 runner and winget rejects the installer before the diff is reached. Removing an `AppsAndFeaturesEntries` `Publisher` line cannot change installer applicability, and these three manifests already carry the "untested, as no test environment was available" agreement. It is deterministic, so a re-run would not change it, and I have not spent one.

The matrix only runs on changed manifests, which is why this has stayed latent — any PR touching an arm32 manifest hits it.

I reverted those three packages in `1064b18` rather than widening this PR into the workflow, so they keep their redundant `Publisher: Microsoft Corporation`. The other 20 dirs and all 8 retargets are unchanged.

Proposed fix for a separate PR, since GitHub has no ARM32 Windows runner to route to — drop `arm` from the matrix in the `Build matrix` step, next to the existing de-duplication:

```powershell
$manifest.Installers | Where-Object { $_.Architecture -ne 'arm' } | ForEach-Object {
```

That skips validation for a target no hosted runner can execute, instead of failing it. Happy to open it if you want it.

---

pl4nty:
Completing the diagnosis — my earlier comment covered only the arm32 jobs, and more failures landed after it. Nine `Installation Validation` jobs fail on `1064b18`, in five unrelated classes. None is caused by this diff.

**Proof, not inference.** I dispatched `validate.yml` against `main` for `OneLogin.DesktopPro 3.1.0.945` — no cleanup commit in the tree ([run 34740876740](https://github.com/pl4nty/winget-extras/actions/runs/34740876740), head `17a655c`). It fails identically:

```
Installer failed with exit code: 1603
Interactive-only install exited with code -1978335159 instead of timing out
MsiInstallProduct installer failed: 1603
Terminating context: 0x8a150049
WDM: Error: System.UriFormatException: Invalid URI: The hostname could not be parsed.
```

Same codes, same installer-log line as on the PR. The OneLogin MSI wants a tenant subdomain it cannot get on a clean runner.

| Package(s) | Failure | Class |
| --- | --- | --- |
| VCRedist arm32 ×3 | `Machine is not compatible with Arm`, `0x8A150010` | No ARM32 runner exists — reverted in `1064b18` |
| Nvidia.Driver.GeForceGameReady 610.62/.74/.88 | installer exit `3858759938` (`0xE0000242`) | No GPU on hosted runners |
| OneLogin.DesktopPro | MSI 1603, `UriFormatException` | Needs tenant config — proven pre-existing above |
| FlexRadioSystems.SmartSDR ×2 | `Interactive-only install exited with code 0 instead of timing out` | `InstallModes: [interactive]` vs `validate.ps1:119`; the MSI completed silently |
| Samsung.NVMeDriver | `Install timed out` after the full 5 min | Driver hangs with no Samsung NVMe device; no `interactive` in `InstallModes`, so `validate.ps1:117` fails it |
| WinDirStat arm64/msix | `HTTP 500 (api.github.com/.../releases/assets/527987446)` then `Cannot find path '...DesktopAppInstaller...msixbundle'` | Transient — died at `validate.ps1:57` before winget was installed |

`AppsAndFeaturesEntries` is read by winget after install, for correlating an installed package back to a manifest. It is never passed to the installer, so removing a redundant `Publisher` or `DisplayName` cannot change an install outcome. Every failure above is either an absent runner capability, a vendor installer needing configuration, an `InstallModes` declaration that disagrees with real installer behaviour, or a GitHub API 500.

The common factor is only that `validate.yml` builds its matrix from *changed* manifests. These packages have not been re-validated since they were added, so a metadata sweep across 28 manifests is the first thing to surface them all at once.

I have not spent a re-run. Only the WinDirStat 500 qualifies for one, and any further push re-runs the whole matrix anyway.

Two observations worth their own issues, which I am not fixing here because one CI run is not enough to characterise an installer:

- `FlexRadioSystems.SmartSDR` declares `InstallModes: [interactive]` but its MSI exited 0 silently.
- `Samsung.NVMeDriver` does not declare `interactive` but hangs past 5 minutes.

Waiting on a maintainer call before touching the remaining packages: revert the un-installable ones from this PR, teach `validate.yml` to skip installation validation for metadata-only diffs, or merge with these documented as pre-existing.

---

PR #1458, opened 2026-09-14
skill: do not preemptively set shard url architecture

pl4nty:
One sentence in `new-package-pr` step 4: leave shard `urls` entries bare, and add an `architecture` override only when a dry run detects it wrong.

Prompted by the `deslop` commit on pl4nty/winget-extras#1418, which replaced an installer object carrying `"architecture": "x86"` with a plain URL string. [Anthelion CONTRIBUTING.md](https://github.com/UnownPlain/anthelion/blob/main/CONTRIBUTING.md) states komac infers architecture from the installer and filename markers, and that overrides must not be added preemptively.

Checklist for Pull Requests

- [ ] Is there a related issue or pull request in [winget-pkgs](https://github.com/microsoft/winget-pkgs)? If so, add the link(s) below.

Manifests

No manifests changed - documentation only, so the manifest checks below do not apply.

- [ ] Have you checked that there aren't other open [pull requests](https://github.com/pl4nty/winget-extras/pulls) for the same manifest update/change?
- [ ] Have you [validated](https://github.com/microsoft/winget-pkgs/blob/master/doc/Authoring.md#validation) your manifest locally with `winget validate --manifest <path>`?
- [ ] Have you tested your manifest locally with `winget install --manifest <path>`?
- [ ] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)?

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MMR1puEV9ppaoow6Q62n5k

---
_Generated by [Claude Code](https://claude.ai/code/session_01MMR1puEV9ppaoow6Q62n5k)_

PR #1460, opened 2026-09-14
fix: shard strategies, release notes, redundant architectures

pl4nty:
A sweep of all 403 shards against Anthelion's [CONTRIBUTING.md](https://github.com/UnownPlain/anthelion/blob/main/CONTRIBUTING.md), for three things: a higher-preference strategy, a script shard that could be JSON, and a missing `releaseNotes`. Then architecture overrides, and the `GROQ_API_KEY` wiring that release notes depend on.

## Strategies

Each of these scraped a source that a higher-preference strategy already parses.

| Shard | Was | Now |
| --- | --- | --- |
| `Microsoft.Aptos.Font` | `page-match` | `ms-download-center` |
| `Microsoft.EquationEditor.Font.Font` | `page-match` | `ms-download-center` |
| `Microsoft.FluentCalibri.Font` | `page-match` | `ms-download-center` |
| `Microsoft.365CompanionApps` | `page-match` | `appinstaller` |
| `Microsoft.SurfaceITToolkit` | `page-match` | `appinstaller` |
| `Microsoft.VisualStudioCode` | `page-match` | `json` |
| `Bluegrams.ScreenRuler` | `sort-versions` | `sourceforge` |

The three font shards regexed the `__DLCDetails__` blob on a Download Center page, which is what `ms-download-center` parses; `Microsoft.FluentCalibri.Font` keeps a name filter because its page also lists an MSI, and the two whose file names contain spaces keep their percent-encoded `urls` because the page publishes them unencoded. Both App Installer packages publish an x64 and an arm64 feed at the same version, so `appinstaller` takes both and their `MainPackage` `Uri` values replace the templated URLs. `Microsoft.VisualStudioCode` regexed a JSON API response; `json` reads `productVersion` and reaches the commit hash as `{data.version}`.

## Release notes

`Composer.WindowsSetup` and `Insecure.Npcap` detect versions from a GitHub release but download from `getcomposer.org` and `npcap.com`, so komac never sees a GitHub URL to resolve notes from. An explicit `{ "source": "github" }` reuses the shard's `owner`, `repo` and resolved tag; cleanup defaults to false for GitHub sources, so it needs no key. Verified: both shards' current manifests have no `ReleaseNotes`, and the dry run produces both `ReleaseNotes` and `ReleaseNotesUrl`.

`GROQ_API_KEY` is now passed to the Update Packages step. Without it the cleanup call returns nothing and Anthelion drops both `ReleaseNotes` and `ReleaseNotesUrl`, so every non-GitHub release-notes configuration was inert. The key is deliberately not added to the test workflow, matching anthelion: dry runs should not spend quota, and secrets are not available to pull requests from forks.

Two shards already fetch a changelog for version detection and now reuse it: `BegoniaHoldings.HopToDesk` (one line per version in 4.6 KB, no limit needed) and `winitor.pestudio` (20 KB, newest first after a preamble, so `characterLimit` 5000).

## Architecture overrides

Sixteen overrides removed, per the rule added in #1458: leave a URL bare and let komac detect the architecture.

Kept, because the override contradicts the installer rather than restating it, and removing it would relabel a shipped package:

- The twelve `Microsoft.DotNet.*.x64` and `.x86` shards label the x64 installer `arm64` and the x86 installer `neutral`, which is what makes the x64 runtime installable on an arm64 host and the x86 runtime installable anywhere.
- `Mozilla.MozillaBuild`, `SeriousBit.NetBalancer` and both `WingFTPSoftware` shards pin x64 while their installer stub is a 32-bit PE.
- `ModMii.ModMii` pins x64 while `ModMii.exe` is 32-bit, and `NirSoft.NirCmd` pins x86 on an archive whose manifest records `neutral`. Both look wrong today, independent of this pull request.
- `Fortinet.FortiClientVPN` could not be checked.

`Rosalie241.RMG` kept one of its two: the dry run detects the Inno installer as x64 unaided, but returns `neutral` for the portable zip, where the manifest records x64.

## Validation

Every changed shard was dry-run and its generated `Architecture` values compared against the shipped manifests, which is how the RMG regression was caught — it was green CI, not a failure.

`Test / Test` failed once on an earlier commit. Each of the twelve shards in that batch then passed individually or in small groups, and the identical commit passed on re-run, so that failure was a flake rather than a bad change.

Release-notes output cannot be verified here: the key is only in the update workflow by design, so a dry run shows no `ReleaseNotes` for the two new plain-text configs. The first real check is the next scheduled update for either package.

---

Checklist for Pull Requests

  - No related winget-pkgs issue; this changes shards and one workflow only.

Manifests

- [ ] Have you checked that there aren't other open [pull requests](https://github.com/pl4nty/winget-extras/pulls) for the same manifest update/change?
- [ ] Does your manifest conform to the [1.28 schema](https://github.com/denelon/winget-pkgs/tree/docs/manifest-schema-1.28.0/doc/manifest/schema/1.28.0)?

No manifests are touched by this pull request.

---

https://claude.ai/code/session_01EPw81MwaPKA4cVNUmxX692


pl4nty:
`Lint Project / Lint Project` is red, but not on anything this pull request changes.

The failing step is `Check Manifests`. `Check Formatting`, `Lint & Type check` and `Test Manifest Validation` all pass.

```
warning[repository/upstream-versions] --> manifests/k/k0sproject/k0sctl/0.33.0/k0sproject.k0sctl.yaml:5:1
  PackageVersion 0.33.0 already exists in microsoft/winget-pkgs
error: script "manifests:check" exited with code 1
```

`bun manifests:check --deny-warnings` scans the whole tree, so a package going upstream turns the check red on every open pull request. This branch touches only `shards/json/*.json` and no manifests at all.

#1459 already removes `k0sproject.k0sctl`, so it is the fix. I could not port that removal into this branch: deleting the files was blocked by the sandbox this session runs in.

No push is needed anyway. CI runs on `pull_request`, so the checkout is the merge ref. Once #1459 lands on `main`, re-running this job picks up the removal and `Check Manifests` should pass without a commit here. I have not spent a re-run yet, because the cause is deterministic and a re-run before #1459 merges would fail identically.

I will keep watching this pull request.

---

Issue #1161, opened 2026-09-04
[Package Request]: Zwift.Zwift.26H1+


DandelionSprout:
### How can we help?

I would like someone else to build the manifest.

### Please read and ensure the following

- [x] The installer meets the above requirements

### Have you submitted to winget-pkgs?

Not applicable; see "Notes from me" below.

### Please provide the following information

Download Page Url: https://www.zwift.com/download
Publisher: Zwift, LLC
Package Name: Zwift.Zwift.26H1+
Description: The app for riding, running, and training with your community. This package's app aims to be identical to that of the winget-pkgs «Zwift.Zwift», except for replacing one dependency to make it work on Windows 11 ≥26H1.
Package Version: 1.1.18
Installer Url: https://cdn.zwift.com/app/ZwiftSetup.exe
PackageDependencies: Microsoft.DotNet.Framework.Runtime.3
MinimumOSVersion: 10.0.28000.0

### Notes from me
In a case of infinite """wisdom""" (/heavy-sarcasm), winget-pkgs have locked down the entire `Microsoft.DotNet` namespace to outsiders (https://github.com/microsoft/winget-pkgs/pull/429144; https://github.com/microsoft/winget-pkgs/pull/426926), making it even less likely than before that winget-pkgs would add `Microsoft.DotNet.Framework.Runtime.3` in time for when 26H2 x64 hits general availability in 1 or 2 months.

Therefore, apps in winget-pkgs that currently rely on the `NetFx3` dependency, would likely need to be added to `winget-extras` where the `NetFx3` dependency can be replaced by `Microsoft.DotNet.Framework.Runtime.3`. Zwift is one of the highest-profile apps among the 20-ish affected apps, and is therefore a good starting point.

The InstallerUrl is static, and shards should be based on that it's static.

pl4nty:
Do you have a 26H2 VM to test these on? I wonder how many apps actually require `NetFx3` with their latest version, or have just inherited it from previous versions. I'd also prefer that the IDs match upstream ie `Zwift.Zwift`

DandelionSprout:
* *"Do you have a 26H2 VM to test these on?"*

Not yet.

* *"I wonder how many apps actually require NetFx3 with their latest version (…)"*

A fair few affected apps have not received updates in many years, and thus very likely still depend on it. `KevinOttalini.SimpleMTUTest` and `LargeAddressAware.laa` are the examples that come first to my mind.

[...]

PR #1492, opened 2026-09-17
Update version: NRaas.Packer version 1.0.0.7

DandelionSprout:
Checklist for Pull Requests

  - Relates to microsoft/winget-pkgs#[Issue Number]

Manifests

---
* Surely there's no way this will actually work, but it wouldn't be the strangest thing I've succeeded with. Nevertheless, it's strange enough that no one would believe me without seeing the changed `installer.yaml` first.


DandelionSprout:
I can't believe that worked. Okay, so here's what this PR does:

1) Since the app relies on NetFx3 as a dependency, and inspired by the circumstances around #1161, the manifest now has one installer that states MinimumOSVersion and uses `Microsoft.DotNet.Framework.Runtime.3`, and one installer that does not state a MinimumOSVersion and uses `NetFx3`.
2) Since Winget (The CLI tool) tends to completely break when faced with the idea of 2 installers that have both the same arch and the same installer type, the MinimumOSVersion installer is set to "Neutral", while the non-MinimumOSVersion installer continues to be set as "x86".
3) How well this will work for ≤25H2 users in effect, remains to be seen, since it's not something that can be tested for without it being in a Winget source.

PR #1521, opened 2026-09-19
New package: Microsoft.Office version 16.0.20326.20112

UnownPlain:
Blocked upstream on a wavier for Manifest-NetworkAddressInSwitch


UnownPlain:
The timeout might need to be longer

pl4nty:
I don't really want to bump the timeout, because it'll increase the wait time for interactive-only installers. I'd be glad if you can think of a better solution though

UnownPlain:
Maybe waviers once we have those implemented?

